PR #36, opened 2015-10-11
Applicative

la-yumba:
Added extension methods to ```Option``` so that it can be used as an applicative functor.

Typical usage:

```
Func<int, int, int> add = (a, b) => a + b;

var result = Some(add)
    .Apply(Some(3))
    .Apply(Some(4));

// result => Some(7)
```

Tests in ```LanguageExt.Applicative.OptionApply``` show usage


louthy:
Hi @la-yumba - Thanks for the pull-request, and for putting the effort into supporting language-ext, however there are a number of reasons I can't currently accept this pull-request:

1. `Apply` is only provided for one of the monadic types.  The main thrust of language-ext is to attempt to create a complete API across all monads (so `OptionUnsafe`, `Either`, `EitherUnsafe`, `Try`, `TryOption`, `Reader`, `Writer`, `State`...).  So I prefer to only add features if they're going to be across all monads, or at the very least the core set: `Option`, `OptionUnsafe`, `Either`, `EitherUnsafe`, `Try`, `TryOption`

2. I also try to maintain a set of functional mirror functions that do the same job as the fluent API, i.e.
```C#
    var result = apply(apply(Some(add),Some(3)),Some(4));
```
It would also make sense to add variants that make the functional API work better:
```C#
    var result = apply(Some(add),Some(3),Some(4));
```
3. There is no need for a separate project file for Applicatives or their unit tests.  I think it's fine that they're in LangaugeExt.Core and LanguageExt.Tests.

4. There's unchecked access to the `Value` property of the `Option` type which is a bug:
https://github.com/louthy/language-ext/pull/36/files#diff-121a151d32da28ae59ac9f9e822ee3c0R14

If you can deal with the issues raised above then I will happily accept the changes.  If not, let me know, as it's something I will probably do at some point also.  I'd just rather not double up the work.

la-yumba:
Hi Paul,

yes, I agree with your comments... this first commit was mostly to get some
initial feedback and see if you were interested!

I can make the last 2 changes you requested right away, but the first 2
will be piece-meal...

maybe you could create a branch on your repo for this work, that we can
both contribute to, and then merge to your master branch when you think
it's ready?


Enrico Buonanno
---
email: enrico.buonanno@gmail.com
skype ID: enrico.buonanno
Tel: +34 632 333 865

louthy:
> maybe you could create a branch on your repo for this work, that we can
both contribute to, and then merge to your master branch when you think
it's ready?

Good plan, here you go:  https://github.com/louthy/language-ext/tree/Applicative

PR #38, opened 2015-10-15
Applicative

la-yumba:
Added extension methods to ```Option``` so that it can be used as an applicative functor.

Typical usage:

```
Func<int, int, int> add = (a, b) => a + b;

var result = Some(add)
    .Apply(Some(3))
    .Apply(Some(4));

// result => Some(7)
```

Tests in ```LanguageExt.Applicative.OptionApply``` show usage


louthy:
@la-yumba That's great, thanks :)

Can you also please make sure you comment each function.  It's been a mammoth job going through all the existing code documenting everything.  So reducing that ongoing workload is definitely something I want to achieve. 

Thanks again.

louthy:
@la-yumba I have now added documentation as well as applicative support for `Option`, `Either` and `EitherUnsafe`

PR #43, opened 2015-10-16
Productable and Divisible support

louthy:


LanguageExt.Core/TryOption.cs
```diff
@@ -144,6 +144,24 @@ public static class __TryOptionExt
         return TypeDesc.Subtract(lhsRes.Value, rhsRes.Value, TypeDesc<T>.Default);
     };
 
+    public static TryOption<T> Product<T>(this TryOption<T> lhs, TryOption<T> rhs) => () =>
+    {
+        var lhsRes = lhs.Try();
+        if (lhsRes.IsFaulted || lhsRes.Value.IsNone) return lhsRes;
+        var rhsRes = rhs.Try();
+        if (rhsRes.IsFaulted || rhsRes.Value.IsNone) return lhsRes;
```
IDisposable: shouldn't this be return rhsRes;

LanguageExt.Core/TryOption.cs
```diff
@@ -144,6 +144,24 @@ public static class __TryOptionExt
...
+        return TypeDesc.Product(lhsRes.Value, rhsRes.Value, TypeDesc<T>.Default);
+    };
+
+    public static TryOption<T> Divide<T>(this TryOption<T> lhs, TryOption<T> rhs) => () =>
+    {
+        var lhsRes = lhs.Try();
+        if (lhsRes.IsFaulted || lhsRes.Value.IsNone) return lhsRes;
+        var rhsRes = rhs.Try();
+        if (rhsRes.IsFaulted || rhsRes.Value.IsNone) return lhsRes;
```
IDisposable: shouldn't this be return rhsRes;

LanguageExt.Core/TypeDesc.cs
```diff
@@ -162,9 +192,58 @@ private static object SubtractNumeric(object lhs, object rhs, TypeDesc desc)
...
+                        case ValueTypeDesc.SByte: return (sbyte)lhs * (sbyte)rhs;
+                        case ValueTypeDesc.Char: return (char)lhs * (char)rhs;
+                        case ValueTypeDesc.UInt: return (uint)lhs * (uint)rhs;
+                        case ValueTypeDesc.ULong: return (ulong)lhs * (ulong)rhs;
+                        case ValueTypeDesc.UShort: return (ushort)lhs * (ushort)rhs;
+                        case ValueTypeDesc.Byte: return (byte)lhs * (byte)rhs;
+                        case ValueTypeDesc.Float: return (float)lhs * (float)rhs;
+                        case ValueTypeDesc.Double: return (double)lhs * (double)rhs;
+                        case ValueTypeDesc.Decimal: return (decimal)lhs * (decimal)rhs;
+                        default: return lhs;
```
IDisposable: Perhaps a throw would be more reasonable? Seems arbitrary.

LanguageExt.Core/TypeDesc.cs
```diff
@@ -162,9 +192,58 @@ private static object SubtractNumeric(object lhs, object rhs, TypeDesc desc)
...
+                        case ValueTypeDesc.SByte: return (sbyte)lhs / (sbyte)rhs;
+                        case ValueTypeDesc.Char: return (char)lhs / (char)rhs;
+                        case ValueTypeDesc.UInt: return (uint)lhs / (uint)rhs;
+                        case ValueTypeDesc.ULong: return (ulong)lhs / (ulong)rhs;
+                        case ValueTypeDesc.UShort: return (ushort)lhs / (ushort)rhs;
+                        case ValueTypeDesc.Byte: return (byte)lhs / (byte)rhs;
+                        case ValueTypeDesc.Float: return (float)lhs / (float)rhs;
+                        case ValueTypeDesc.Double: return (double)lhs / (double)rhs;
+                        case ValueTypeDesc.Decimal: return (decimal)lhs / (decimal)rhs;
+                        default: return lhs;
```
IDisposable: Perhaps a throw would be more reasonable? Seems arbitrary.

LanguageExt.Core/TryOption.cs
Therzok: Good catch!

louthy:
> shouldn't this be return rhsRes;

Yes, fixed.

> Perhaps a throw would be more reasonable? Seems arbitrary. Maybe even returning a None state?

'In theory' it's impossible to get there.  However, I'm also aware of the first law of programming: If it can't possibly happen, it will.  So I've put the relevant throws.

Thanks :)

PR #48, opened 2015-10-23
Mostly clean up warnings plus one fix

Jagged:
See commit log for details

### DIFF
--- a/LanguageExt.Core/Disposable.cs
+++ b/LanguageExt.Core/Disposable.cs
@@ -22,11 +22,8 @@ namespace LanguageExt
 
         public void Dispose()
         {
-            if (Value != null)
-            {
-                Value.Dispose();
-                Value = null;
-            }
+            Value?.Dispose();
+            Value = null;
         }
     }
 }
--- a/LanguageExt.Core/Disposable.cs
+++ b/LanguageExt.Core/Disposable.cs
@@ -11,7 +11,7 @@ namespace LanguageExt
     }
 
     public class LinqDisposable<T> : ILinqDisposable
-        where T : IDisposable
+        where T : class, IDisposable
     {
         public readonly T Value;
 
--- a/LanguageExt.Core/Disposable.cs
+++ b/LanguageExt.Core/Disposable.cs
@@ -13,7 +13,7 @@ namespace LanguageExt
     public class LinqDisposable<T> : ILinqDisposable
         where T : class, IDisposable
     {
-        public readonly T Value;
+        public T Value { get; private set; }
 
         internal LinqDisposable(T value)
         {
@@ -22,10 +22,10 @@ namespace LanguageExt
 
         public void Dispose()
         {
-            var value = Value;
-            if (value != null)
+            if (Value != null)
             {
-                value.Dispose();
+                Value.Dispose();
+                Value = null;
             }
         }
     }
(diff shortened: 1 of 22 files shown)


louthy:
Hi @Jagged - First of all, thanks so much for contributing to the project.  A second pair of eyes is definitely appreciated!  I am unfortunately going to have to reject your pull-request in its current state however.  It is mostly all good, but there are a few things that (if you could) I'd like you to improve:

Some of the `Dispose()` modifications you've made are unsafe.  

This can fail if called from multiple threads:
```C#
    public void Dispose() 
    {
        if( val != null )
        {
            //  <--- If another thread calls this Dispose method then val may == null here
            //       Which would throw a NullReferenceException
            val.Dispose();
            val = null;
        }
    }
```
So you should do this:
```C#
    var localVal = val;
    if( localVal != null )
    {
        localVal.Dispose();
        val = null;
    }
```
It's up to the `IDisposable` class to deal with double entry.  So if it was a `FileStream` for example it may use locks where the LangExt disposables don't need them.

I don't use that pattern on the Enumerators because mostly you know they'll be called via a `foreach` and will therefore only be disposed once and Enumerators aren't normally passed between threads.

Disposable.cs

* There is a bug in the `Dispose`, but it should be fixed thus:
```C#
    public void Dispose()
    {
        var value = Value;
        if (value != null)
        {
            value.Dispose();
            Value = null;
        }
    }
```
* Don't change `readonly` fields to properties - please revert.

Actor.cs

* I think it's unnecessary to protect against double disposal; it should be the state's `Dispose` method that deals with it.  So there's no need for `isDisposed`.  Also removing subscriptions twice won't hurt.  In the unlikely event that `Dispose` gets called multiple times I can deal with the hit.
* Obviously the bug fix is welcome though, so I think the `Dispose` method should be reverted and the `DisposeState` method be left with your change.

ActorInboxDual.cs - Revert

ActorInboxLocal.cs - Should be:
```C#
    var ts = tokenSource;
    ts?.Dispose();
    tokenSource = null;
```
ActorInboxRemote.cs - Should be:
```C#
    var ts = tokenSource;
    ts?.Cancel();
    ts?.Dispose();
    tokenSource = null;
```
RedisClusterImpl.cs

* Please revert - I know it's not great practice to call `Dispose` from a finaliser, but in reality I've seen a cluster survive across an ASP.NET app-pool recycle which can cause major problems (especially whilst an app is in development and you're dropping out of debugging sessions).  So I want to make absolutely 100% sure that the cluster and connections are shut-down no questions asked.  It's brutal, but necessary.

List.cs - Fine
Lst.cs - Fine
MapKV.cs - Fine
ObjectPool.cs - Fine
Prelude_Obsolete.cs - Fine (if unnecessary due to the methods being obsolete)
Prelude_Use.cs - Fine
RWS.cs - Fine
Reader.cs - Fine
SetT.cs - Fine
State.cs - Fine
Try.cs - Fine
TryOption.cs - Fine
Writer.cs - Fine
Prelude_Logging.cs - Fine
LanguageExt.ProcessJS.csproj - Fine

Again, thanks for the contribution.  If you disagree with anything I've written above please let me know and we can discuss it further.

Paul

Jagged:
Hi Paul, thanks for your feedback.  I'll go thru things and amend to
protect against threads, I suspected you might pull me up on that!  The new
?. operator is quite useful here!

*WILL MOOAR*
Enterprise Software Consultant, Pebkac Ltd, 027 6968306,
will.mooar@pebkac.nz
<http://pebkac.nz/>

Jagged:
I think I've completed all you've asked, though with a little artistic license. =)  I noticed that ActorInboxLocal wasn't Cancelling the tokenSource, so added that.

louthy:
> I think I've completed all you've asked, though with a little artistic license. =) I noticed that ActorInboxLocal wasn't Cancelling the tokenSource, so added that.

Thanks Will, that all looks great!

PR #51, opened 2015-10-24
Minor code refactoring

Jagged:
see commits


louthy:
@Jagged - Good stuff!  Does that Refactoring Extension do all this automatically?  Or are you going through manually?

Jagged:
It suggests a bunch of things, and then I selectively pick thru them so I don't mess up the style of your coding.  For instance you often use ternary expressions where a simpler binary expression could do, but I get that you've written it that way for readability, so left them alone.    ... for now ... =P

But there's many things it can do, and if you like what it's offering then you can, with a couple of clicks, change one, some or all occurrences.

It's rather powerful!  But be cautious, I have seen it generate incorrect code that still compiled.  That's rare though.  I manually review the commits just in case.  Still much faster than by hand!

ps It's free!  And doesn't seem to affect VS speed very noticeably unlike some popular commercial tools that do similar.

louthy:
> For instance you often use ternary expressions where a simpler binary expression could do
Yeah this is actually a blind spot for me.  I tend to fall into the ternary operator a lot.

Cool, I've installed it now. Thanks for the heads up :)

PR #52, opened 2015-10-24
Harden Some so it will fail instead of being treated as None when uninitialised

Jagged:
per commits, including tests

### DIFF
--- a/LanguageExt.Core/Some.cs
+++ b/LanguageExt.Core/Some.cs
@@ -26,6 +26,11 @@ namespace LanguageExt
         }
     }
 
+    /// <summary>
+    /// Holds a value that is guaranteed not to be null.
+    /// Use MatchUntyped or Value to access the value.
+    /// </summary>
+    /// <typeparam name="T"></typeparam>
     [TypeConverter(typeof(SomeTypeConverter))]
     public struct Some<T> : IOptional
     {
@@ -60,27 +65,33 @@ namespace LanguageExt
             value.Value;
 
         public override string ToString() =>
-            Value.ToString();
+            initialised
+                ? value.ToString()
+                : "(uninitialised)";        // ToString is not allowed to throw exceptions
 
         public override int GetHashCode() =>
-            Value.GetHashCode();
+            initialised
+                ? value.GetHashCode()
+                : 0;                        // GetHashCode is not allowed to throw exceptions
 
         public override bool Equals(object obj) =>
-            Value.Equals(obj);
+            initialised
+                ? value.Equals(obj)
+                : false;                    // Equals is not allowed to throw exceptions
 
         public bool IsSome =>
-            initialised;
+            CheckInitialised(true);
 
         public bool IsNone =>
-            !initialised;
+            CheckInitialised(false);
 
         public R MatchUntyped<R>(Func<object, R> Some, Func<R> None) =>
             IsSome
                 ? Some(value)
-                : None();
+                : raise<R>(new InvalidOperationException("should never be None"));
 
         public Type GetUnderlyingType() =>
-            typeof(T);
+            CheckInitialised(typeof(T));
     }
 
     public static class Some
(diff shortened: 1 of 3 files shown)


louthy:
Hi Will,

There are problems here:

Some.cs - 

The documentation:

> Holds a value that is guaranteed not to be null.

It's not guaranteed to not be null.  So this is misleading.  For example:
```C#
    // This holds 100 Some(null)
    var x = new Some<object>[100];  
```
> Use MatchUntyped or Value to access the value.

This is slightly misleading.  `MatchUntyped` would be a rarely used method (for example when serialising).

Access to the value would mostly be done via implicit conversion:
```C#
	Some<MyObj> x;
	MyObj y = x;
```
Or by `Value` (as your comment states):
```C#
	MyObj y = x.Value;
```
So I think re-wording those comments to remove the reference to `MatchUntyped` and put the focus on implicit conversion would be better.

Please remove the comments below - they're commit notes rather than code documentation, the code is self explanatory: 
```
    // ToString is not allowed to throw exceptions
    // GetHashCode is not allowed to throw exceptions
    // Equals is not allowed to throw exceptions
```
The other changes to `ToString`, `GetHashCode` and `Equals` are fine though.

The changes to `IsSome` and `IsNone` are bugs.  By calling `CheckInitialised` you remove the chance that the programmer can interrogate the state of the object because it will throw an exception if `initialised == false`.

The change to `MatchUntyped` is also a bug.  Any call to `match` throughout the library is there to allow the programmer to deal with failure/invalid states.

`GetUnderlyingType` also doesn't need protection with `CheckInitialised`

Thanks,

Paul

Jagged:
The changes to IsSome, IsNone and MatchUntyped were supporting hardening Some as much as possible, to fail fast if nulls or initialised snuck in.  

However I understand the changes aren't what you're wanting for Some, so I'll drop them and rethink my approach.  Cheers

louthy:
>The changes to IsSome, IsNone and MatchUntyped were supporting hardening Some as much as possible, to fail fast if nulls or initialised snuck in.

Yep, I appreciate the reasoning; it would make `Some` inconsistent with the rest of the library however and it wouldn't be honouring the `IOptional` interface.  You could argue that it shouldn't implement `IOptional`, but seeing as the value _is_ optional, I think it's valid to give those tools to the programmer.

One example of where I have personally found this useful is in a text-interpolation library I have written.  It uses reflection to dot into objects that are wrapped in braces, i.e. `{Foo.Bar}`.  As well as checking for `null` it also checks for optional values and allows alternatives to be selected when `IsNone == true`.  i.e. `{Foo.Bar|Foo.Alt}`.  It needs `IsNone`, `IsSome`, and `MatchUntyped` to handle failure in the same graceful way that the `Option` and `OptionUnsafe` types do.  

There are a few types in this library that can get into an uninitialised state.  I have co-opted the Haskell term `bottom`; which means a computation that never completes [1] - but in our world it means you could get an exception on use.  There are occasions, however, that you wouldn't get an exception: it's mostly to facilitate expression based programming (either in LINQ or through the fluent interfaces).

For example `Either<L,R>` can be in three states, not just the two that most people understand as `Left` and `Right`.  It also has `Bottom`.  That can mean 'uninitialised' if declared in an array or as a member variable, or in this scenario:
```C#
    Either<string, int> a = 100;   // Right(100)

    var y = from x in a
            where x < 100
            select x;

    Assert.True( y.IsBottom );
```
If the predicate fails and the `Either` is filtered out, then it still needs a representation as an `Either`.  `IEnumerable` can do it by removing the item from the collection, but an `Either` represents one value always:  `Right` or `Left`.  If it used `default(L)` and put the instance into a `Left` state then you could have `Left(null)`, which wouldn't be any use at all.  So that's where the third state `Bottom` comes in.

Some operations will throw a `BottomException`, others will not.  Mostly if you're trying to extract the value then you will get an exception.  However operations like `Map`, `Filter`, `Fold`, `Iter`, `Count`, `Sum` etc. will continue because they can treat the `Either` as 'empty'.  
```C#
    // y is in a bottom state
    var z = y.Map( a => a + 1 ).Sum();
    Assert.True( z == 0 );
```
This allows expressions to continue and just return an `Either` in the bottom state at the end rather than blowing up with an exception.  

Anyway, I hope that gives an insight into the potential validity of uninitialised state.  

[1] https://wiki.haskell.org/Bottom

PR #56, opened 2015-10-30
Tightened null checks in EitherT.

Jagged:
Also added ~~`Bottom() helper`~~ `static readonly Bottom` so internal classes can create Bottom instance in same easy way as Left and Right helpers.

And tidied up some remaining IProductable bits to IMultiplicable, and anything similar.

### DIFF
--- a/LanguageExt.Core/Either.cs
+++ b/LanguageExt.Core/Either.cs
@@ -38,18 +38,26 @@ namespace LanguageExt
         IProductable<Either<L, R>>,
         IDivisible<Either<L, R>>
     {
+        public readonly static Either<L, R> Bottom = new Either<L, R>();
+
         readonly R right;
         readonly L left;
 
-        private Either(Some<R> right)
+        private Either(R right)
         {
+            if(right == null)
+                throw new ValueIsNullException();
+
             this.State = EitherState.IsRight;
             this.right = right;
             this.left = default(L);
         }
 
-        private Either(Some<L> left)
+        private Either(L left)
         {
+            if (left == null)
+                throw new ValueIsNullException();
+
             this.State = EitherState.IsLeft;
             this.right = default(R);
             this.left = left;
@@ -67,12 +75,14 @@ namespace LanguageExt
         /// <summary>
         /// Is the Either in a Right state?
         /// </summary>
+        /// <exception cref="BottomException">EitherT state is Bottom</exception>
         public bool IsRight =>
             CheckInitialised(State == EitherState.IsRight);
 
         /// <summary>
         /// Is the Either in a Left state?
         /// </summary>
+        /// <exception cref="BottomException">EitherT state is Bottom</exception>
         public bool IsLeft =>
             CheckInitialised(State == EitherState.IsLeft);
 
@@ -95,7 +105,8 @@ namespace LanguageExt
         /// <summary>
         /// Implicit conversion operator from R to Either R L
         /// </summary>
-        /// <param name="value">Value</param>
+        /// <param name="value">Value, must be null.</param>
+        /// <exception cref="ValueIsNullException">Value is null</exception>
         public static implicit operator Either<L, R>(R value) =>
             value == null
                 ? raise<Either<L, R>>(new ValueIsNullException())
@@ -104,7 +115,8 @@ namespace LanguageExt
         /// <summary>
         /// Implicit conversion operator from L to Either R L
         /// </summary>
-        /// <param name="value">Value</param>
+        /// <param name="value">Value, must be null.</param>
+        /// <exception cref="ValueIsNullException">Value is null</exception>
         public static implicit operator Either<L, R>(L value) =>
             value == null
                 ? raise<Either<L, R>>(new ValueIsNullException())
@@ -521,14 +533,11 @@ namespace LanguageExt
                 ? raise<U>(new BottomException("Either"))
                 : value;
 
-        internal static Either<L, R> Right(Some<R> value) =>
+        internal static Either<L, R> Right(R value) =>
             new Either<L, R>(value);
 
-        internal static Either<L, R> Left(Some<L> value) =>
+        internal static Either<L, R> Left(L value) =>
             new Either<L, R>(value);
-
-        internal static Either<L, R> Bottom() =>
-            new Either<L, R>();
 
         internal R RightValue =>
             CheckInitialised(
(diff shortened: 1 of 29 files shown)


louthy:
Rejecting for the following reasons:

1. Don't use `Some<T>` for all of the reasons previously discussed.  
2. `Either<L, Res>.Bottom()` would be better as `Either<L, Res>.Default` where `Default` is `static readonly Either<L,R> Default`.  That removes the allocation overhead of the current solution and your replacement.

Jagged:
Changed per feedback, though called Default "Bottom" instead for clarity, which removed need for Bottom() function.  Shame we can't use SomeT, even for internal methods, it made things very easy to add null protection.

Jagged:
And FYI, structs do not allocate memory from the heap, however their members might.  There are many articles on the web regarding this topic.  What this means is that whenever you pass a variable containing a struct, it copies the whole struct.  That's why MS has recommendations to keep struct size small.  Great for SomeT, OptionT, EitherT etc.

The assembly below shows that "new" is faster than "static EitherT Default" as the new only needs to zero the memory for the struct, whereas using Default needs to copy the Default values to the new struct.   You'll note that there's no memory allocation.

            var x = new Either<string, int>();
00007FFEC8AF4A1C 33 C0                xor         eax,eax  
00007FFEC8AF4A1E 48 8D 55 28          lea         rdx,[rbp+28h]  
00007FFEC8AF4A22 66 0F 57 C0          xorpd       xmm0,xmm0  
00007FFEC8AF4A26 F3 0F 7F 02          movdqu      xmmword ptr [rdx],xmm0  

            var x = Either<string, int>.Bottom;
00007FFEC8AE496C 48 B8 28 89 78 AA E5 00 00 00 mov         rax,0E5AA788928h  
00007FFEC8AE4976 48 8B 00             mov         rax,qword ptr [rax]  
00007FFEC8AE4979 48 83 C0 08          add         rax,8  
00007FFEC8AE497D F3 0F 6F 00          movdqu      xmm0,xmmword ptr [rax]  
00007FFEC8AE4981 F3 0F 7F 45 28       movdqu      xmmword ptr [rbp+28h],xmm0  

In short, struct is great as it doesn't have the overheads of heap allocation, only takes up as much memory as the fields, can't be null, and perfect for immutables.  However it doesn't have a default ctor (yet - c#7 maybe), and as it always copies the struct everywhere it is inefficient for larger structs.  I think 16 bytes was MS' upper suggested limit.  Oh and doesn't support all the class semantics.  Still, it has it's uses!

louthy:
Please stop arguing with me regarding structs and allocation, I know exactly how they work.  Please take a look at Eric Lippert's comments on this:  http://blogs.msdn.com/b/ericlippert/archive/2009/04/27/the-stack-is-an-implementation-detail.aspx

I will review the PR tomorrow.

louthy:
And this too: http://blogs.msdn.com/b/ericlippert/archive/2010/09/30/the-truth-about-value-types.aspx

Jagged:
Thanks for responding.  

I'm sorry, it was NEVER meant to be taken as arguing.  I'd backed off re SomeT as the commits show, and then in good faith reworked the code and also helped finish off the Productable to Multiplicable rename.  ``I hope you can see it's the actions of someone who is trying to help.``  Of course I'll question, and sometimes accidentally go in the wrong direction, and that's where I look forward to your constructive feedback.

However "Please stop arguing with me" isn't constructive, it's making it personal.  I was just trying to understand what you were saying vs what the compiler does - there is definitely a disconnect.  But I'm humble enough to know I can't know everything, and I'm getting ready to eat my hat!

I will have a good read.  Thanks for the links.

louthy:
Hi Will, thanks for understanding.  It's not personal, but I had already made my position clear.  I have a limited amount of free time to do this project, and going over the same point multiple times can suck that time away.  

> struct types are stack allocated unless they're part of another class. They are not heap allocated by themselves. Yes it's an implementation detail, but it's still a fact for today's compilers.

That still isn't correct.  It's more nuanced.  If they're captured as part of a closure (which is an anonymous class) then there will be allocation.  A lot of this library is based around the use of closures.  So using `Some<T>` would without a doubt do more allocations than is necessary - I have to be careful to not add a burden where it has little benefit.  Yes it means manually checking for `null`, and that's annoying, but it's a price that's worth paying for improved performance.

louthy:
PR review:

The comment says `Value, must be null`.  Should be `Value, must not be null`
https://github.com/louthy/language-ext/pull/56/files#diff-10ea3cb60b61e03c47de7a19e0abf023R108

Perhaps change the strings of `"Left"`, `"leftMap"`, etc. into `nameof(Left)`, `nameof(leftMap)` etc.

Everything else looks good.

Jagged:
Good spotting, have corrected and made use of nameof, much nicer!

Thanks for explaining where you're coming from, though you raised more points that you cleared!  ("oh no" I hear you say =P  .. though I can imagine other words lol )

``without a doubt do more allocations`` 

If a closure was going to capture a variable then it was already going to have to allocate space via an anonymous object to hold it.  True?

Wrapping the value in a struct (eg SomeT/OptionT) wouldn't change the number of allocations, as they're just an inline wrapper for the contained value.  Also true?

So, um, where's the extra allocation?  
Or maybe I'm just having an extra thick day?

(I'm the kind of person who can't help but ask when something appears not to add up, especially if others are making decisions based on potentially questionable info - it's what I do for a job, as I'm sure you do too.  But hey, I could be wrong!  But I've found it never hurts to question and explore things.  Well, almost never. =) )

ps I'm about to submit a PR to nuke SomeT from high orbit so I'm not tempted to use it in future... =)

louthy:
> If a closure was going to capture a variable then it was already going to have to allocate space via an anonymous object to hold it. True?

I'd suggest reading that second like I posted - especially the last few paragraphs.  It would have to allocate twice, once for the anonymous object and once for the struct member that is allocated on the heap.

Please I'd really like to drop this now.  I am not going to accept `Some` in the project.  The memory allocation was one of the reasons I have given several times already, the other was the lack of covariance in generic structs.  They both add up to problems.  So we're wasting time discussing it, sorry.

> ps I'm about to submit a PR to nuke SomeT from high orbit so I'm not tempted to use it in future... =)

Don't.  People are using this library.  I don't want to fuck them over by dropping features they're using.  `Some<T>` can be useful for certain circumstances.  It's just not general enough for this library's API.

Jagged:
No probs.  This wasn't about SomeT, I'd dropped it ages ago.  I was just wanting to understand where the extra allocations came from for structs in general.

Cheers =)

PR #65, opened 2015-11-08
Some docs for fold methods (wip) and use "isnull(...)"

Jagged:
Heya Paul, I saw you had started using "isnull" to clean up warnings, so I used VS search and replace to complete the job across the code base.  I had a couple of other changes pending too for docs, so they're mixed up in this PR.

Cheers
- Will


louthy:
The changes to the docs where you put "if Right", "if Left", etc. should be removed.  It's still a fold regardless of the state of the underlying monad. Imagine a list with zero items in, if you do a fold on that then you just get your input state back; it is however still a fold, and that applies to `Option` and `Either` types and any type that can have a 'no value' state.

I'm also not sure replacing every `... == null` with `isnull(...)` is such a great idea.  it's obviously worth doing for generic types, but there are a lot of places where you've it used on reference types and would add the overhead of a function call where it's not really necessary.  I could probably put the `AggressiveInlining` attribute on the `isnull` and `notnull` functions however, which would probably solve that issue; but was there a reason you applied it to all null checks?

Jagged:
Evening! 
Re "if left/right" , done. =)
Re "isnull" for ref types, done.  My bad, I had used IDE search and replace and it picked up more than I'd intended.  The only reason to do any of this was to help finish off the work you'd started with the generics.  Hopefully things make more sense now with my latest commits.
- Will

louthy:
> Re "if left/right" , done. =)

It's the same for the other monads too:

https://github.com/louthy/language-ext/pull/65/files#diff-f82b6b3e8bdc33537e9dbbff34b9f63cR604

https://github.com/louthy/language-ext/pull/65/files#diff-d9646d9b4cf5ff61782ddc0f2963e5a2R186

https://github.com/louthy/language-ext/pull/65/files#diff-9063c670a7e6125a3c7c073e64dc00bdR100

https://github.com/louthy/language-ext/pull/65/files#diff-9063c670a7e6125a3c7c073e64dc00bdR100

https://github.com/louthy/language-ext/pull/65/files#diff-ceca91f15305b8784e611d30d8c95851R141

https://github.com/louthy/language-ext/pull/65/files#diff-62b2de3e24de9a5e255d21b4755c4297R536

https://github.com/louthy/language-ext/pull/65/files#diff-3716676838224af20d9e96d45d8a64abR500

Also instead of saying, 'fold the Try result' or 'fold the TryOption result', it would be better to say 'fold the Try value'.  `Try` and `TryOption` being delegates isn't really relevant, and introducing the idea of a result isn't useful.  They're _lazy values_ rather than _delegates that return a value_ whereas the other non-delegate monads are _strict values_.

https://github.com/louthy/language-ext/pull/65/files#diff-ceca91f15305b8784e611d30d8c95851R141

https://github.com/louthy/language-ext/pull/65/files#diff-62b2de3e24de9a5e255d21b4755c4297R536

https://github.com/louthy/language-ext/pull/65/files#diff-3716676838224af20d9e96d45d8a64abR500

Jagged:
Have searched in every way I can think of to clean up the rest of the docs as mentioned.  I hope I got it all.  I had only just started cleaning up the Fold documentation before I got side-tracked, it's still very inconsistent.  I'll have another go when this PR is out of the way.

Jagged:
Morning!  Is there anything I've missed holding up this PR?

louthy:
Hi Will, I'm away at the moment, I'll check when I'm back (Tuesday).

PR #82, opened 2016-03-30
Bigfix - missing tags in the Message Serializer

StanJav:
The strategy system doesn't work properly because of missing tags.


louthy:
@StanJav Thanks Stan :)

PR #83, opened 2016-03-31
Task monad

louthy:
Added extension methods for `Task<T>` so that it can work with the 'higher-kinded type' system.
i.e.

```C#
        Task<Uri> parseUri(string uri) => 
            new Uri(uri).AsTask();

        Task<WebRequest> openConnection(Uri uri) =>
            WebRequest.CreateDefault(uri).AsTask();

        Task<WebResponse> getInputStream(WebRequest req) =>
            req.GetResponseAsync();

        Task<Stream> getSource(WebResponse resp) =>
            resp.GetResponseStream().AsTask();

        Task<IEnumerable<string>> getLines(Stream stream) =>
            readAllLines(stream).ToList().AsEnumerable().AsTask();

        public Task<IEnumerable<string>> getURLContent(string url) =>
            from u in parseUri(url)
            from conn in openConnection(u)
            from result in use(
                getInputStream(conn),
                stream => use(getSource(stream), getLines)
                )
            select result;

        [Fact]
        public void UrlTest()
        {
            getURLContent("http://www.google.com").IterT(Console.WriteLine);
        }
```


LanguageExt.Core/Try.cs
```diff
@@ -706,20 +706,20 @@ public static Try<T> Filter<T>(this Try<T> self, Func<T, bool> Succ, Func<Except
                 var resU = bind(resT.Value).Try();
                 if (resU.IsFaulted)
                 {
-                    if (resT.Value is ILinqDisposable) (resT.Value as ILinqDisposable).Dispose();
+                    //if (resT.Value is ILinqDisposable) (resT.Value as ILinqDisposable).Dispose();
```
IDisposable: We shouldn't check in commented-out code... if it's dead, that's what GIT is for...

PR #90, opened 2016-05-16
Small fixes in README

Hinidu:
Hello! This library looks great!
While reading of README I found some small errors and tried to fix them:
Replaced `ask<V>` by `ask<string>` in the example of cache with static class.
Removed `ProcessId.User` usage because this property doesn't exist.
Removed unnecassary `tell` in example of `ProcessId` API.


louthy:
@Hinidu - thanks!

PR #96, opened 2016-06-13
Add async extensions for Option<T>

danielcweber:
Added __OptionExt.MapAsync and BindAsync.

### DIFF
--- a/LanguageExt.Core/Option.cs
+++ b/LanguageExt.Core/Option.cs
@@ -716,6 +716,11 @@ public static class __OptionExt
             ? Optional(mapper(self.Value))
             : None;
 
+    public static async Task<Option<R>> MapAsync<T, R>(this Option<T> self, Func<T, Task<R>> mapper) =>
+        self.IsSome
+            ? Optional(await mapper(self.Value))
+            : None;
+
     [Pure]
     public static Option<R> Map<T, R>(this Option<T> self, Func<T, R> Some, Func<R> None) =>
         self.IsSome
--- a/LanguageExt.Core/Option.cs
+++ b/LanguageExt.Core/Option.cs
@@ -768,6 +768,11 @@ public static class __OptionExt
             ? Some(self.Value)
             : None();
 
+    public static async Task<Option<R>> BindAsync<T, R>(this Option<T> self, Func<T, Task<Option<R>>> binder) =>
+        self.IsSome
+            ? await binder(self.Value)
+            : None;
+
     [Pure]
     [EditorBrowsable(EditorBrowsableState.Never)]
     public static Option<U> Select<T, U>(this Option<T> self, Func<T, U> map) =>


louthy:
@danielcweber 

Hi Daniel, thanks for this.  However at this time I won't accept.  I like to keep all of the monadic types in-sync, and so adding `BindAsync` an `MapAsync` to `Option` would also require it adding to:

`OptionUnsafe`, `Either`, `EitherUnsafe`, `Try`, `TryOption`, `Lst`, `Set`, `Map`, `Que` (and optionally `RWS`, `State`, `Writer`, `Reader`)

There are also the `Prelude_...` functions with the lower-case static variants (so `bindAsync`, `mapAsync`)

The reason for trying to keep them all in-sync is the nightmare situation that F# got itself in with `List`, `Seq`, `Array`, where each module supported a different set of functions (that's now resolved I believe); and it was massively annoying.  So any new function that comes in needs to be implemented across the board (where possible), to give a consistent and reliable interface.

A final benefit of this is that I can add the functions to the transformer system, so you should be able to do things like `Lst<Option<T>> x;  x.MapAsyncT(...)`

If you're happy to do those that'd be awesome, but yeah, welcome to my pain ;-)

danielcweber:
If you find the addition generally useful, I'll give it a try. Is the list of types you gave exhaustive ?

louthy:
@danielcweber Yep, although I've just noticed that `MatchAsync` isn't implemented for all.  Leave it with me and I'll sort it.

louthy:
@danielcweber I've made a start on this, you can grab the latest from nu-get `1.8.18-beta`

It includes:

`MatchAsync` for `Task<Option<T>>`, `Option<Task<T>`, `Task<OptionUnsafe<T>>`, `Option<TaskUnsafe<T>`, `Try<Task<T>>`, `Task<Try<T>>`, `TryOption<Task<T>>`, `Task<TryOption<T>>`

The following are only done for `Option<Task<T>>`, `Task<Option<T>>`, `OptionUnsafe<Task<T>>`, `Task<OptionUnsafe<T>>`, `Either<L, Task<R>>`, `Task<Either<L, R>>`, `EitherUnsafe<L, Task<R>>`, `Task<EitherUnsafe<L, R>>`.  

* `MapAsync` / `mapAsync` 
* `BindAsync` / `bindAsync`
* `IterAsync` / `iterAsync`
* `FoldAsync` / `foldAsync`
* `ForAllAsync` / `forallAsync`
* `ExistsAsync` / `existsAsync`
* `CountAsync` / `countAsync`
* `SumAsync` / `sumAsync`

I will complete the rest over the next few days, but I figured these would be useful for you now.

danielcweber:
Work great! Makes for some beautiful code. Thanks a lot. If I can contribute to this otherwise, please get back to me.

PR #98, opened 2016-06-20
Fix build with xbuild

jonny-novikov:
Fix build process with xbuild (v4) failed, for example, on Travis.


louthy:
@jonny-novikov Thanks, I'll merge now/

PR #138, opened 2016-09-15
Fixed Option<T>.IfNoneUnsafe samples

Maximusya:
These lines won't compile:
```C#
    string x = ifNoneUnsafe( optional, null );
    string x = optional.IfNoneUnsafe(null);
```

> The call is ambiguous between the following methods or properties: 'LanguageExt.Option<string>.IfNoneUnsafe(System.Func<string>)' and 'LanguageExt.Option<string>.IfNoneUnsafe(string)'

Ofc it should specify a cast (just like in the examples a bit above).

### DIFF
--- a/README.md
+++ b/README.md
@@ -313,7 +313,7 @@ So you can use `matchUnsafe` and `ifNoneUnsafe`:
                             Some: v => v,
                             None: () => null );
 
-    string x = ifNoneUnsafe( optional, null );
+    string x = ifNoneUnsafe( optional, (string)null );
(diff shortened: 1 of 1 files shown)


Therzok:
Better use `default(string)`

Maximusya:
End-user may use whatever she likes, but for the sake of documenting *Unsafe* and *null* are read like "null is unsafe". And btw internal methods have *null* in their names all over the place (CheckNullReturn, isnull etc)

louthy:
@Maximusya Thanks!

PR #139, opened 2016-09-15
Documentation fixes for List/IEnumerable ForAll methods

Maximusya:
The docs incorrectly mentions "Any in LINQ" for ForAll methods


louthy:
@Maximusya Good spot, thanks!

Issue #142, opened 2016-09-22
Deserialize doesn't use same JsonSerializerSettings as Serialize


spyder94:
In several places JsonConvert.SerializeObject is called for message serialization.  In each case ActorSystemConfig.Default.JsonSerializerSettings is passed along with has the TypeNameHandling value set to All.  However, the call to JsonConvert.DeserializeObject does not pass any JsonSerializerSetting.  That causes an exception to occur when a message contains a property that is of a type that is an abstract base class.

I was able to correct this by modifying LanguageExt.Process/ActorSys/Deserialize.cs to invoke JsonConvert.DeserializeObject<T>(string, JsonSerializerSettings) at line 41 after changing the DeserialiseFunc method to find that method instead of the one that only takes the single string parameter.

louthy:
I'm currently away, will review this in a few days.  It sounds like a legitimate bug, so if you want to do a pull request then I'll review when I'm back.

louthy:
@spyder94 This is now released to NuGet as v1.9.3

PR #145, opened 2016-09-26
JsonSerializerSettings for Deserialize

spyder94:
This change makes the deserialization use the ActorSystemConfig.Default.JsonSerializerSettings.  Those settings are used when serializing already.  See issue #142 for more info.


louthy:
@spyder94 Excellent, thanks!

PR #147, opened 2016-09-27
Bug fix for Lst<T>::operator !=

AndreyTsvetkov:
Small fix for the case of nulls. 

The invariant 

> (lst1 == lst2) always gives opposite of (lst1 != lst2)

was not true, not it is true


louthy:
Thanks Andrey :)

Issue #153, opened 2016-10-09
Union Type with functional Match syntax


sdedalus:
I've been experimenting with adding a Union type to language-ext with the following syntax. 

```
            var x = new Union<string, int>(100);
            // the following syntax also works
            // Union<string, int> x = 100;

            string value = x.Match<string, int, string>()
                (a => a)
                (b => b == 100 ? "It's 100." : "It's Not 100.");

            Assert.Equal("It's 100.", value);
```
or
```
            var x = new Union<string, int>("Test");
            string value = x.Match<string, int, string>()
                (a => a)
                (b => b.ToString());

            // the following syntax also works
            //string value = x.Match<string, int, string>(
            //    a => a,
            //    b => b.ToString());

            Assert.Equal("Test", value);
```
I've written the code to handle unions of up to eight types.  
Is there any interest in adding this to language-ext?

louthy:
@sdedalus Hi David, going to close this for now.  Feel free to re-open if you want to discuss further.

PR #154, opened 2016-10-10
Generic Union type with match.  

sdedalus:
Union Type with functional Match syntax mentioned in issue #153.

Discriminated Unions of up to eight types can be created.  Unions support implicit casting to the union type to simplify return statements.  the match syntax sticks as closely to the F# syntax as I could manage in C# and includes both a curried syntax and a multi parameter syntax.

### DIFF
--- a/LanguageExt.Core/Prelude_Union.cs
+++ b/LanguageExt.Core/Prelude_Union.cs
@@ -4,24 +4,34 @@
 
 	public static partial class Prelude
 	{
-		public static IUnion<B> Bind<A, B>(this IUnion<A> a, Func<A, IUnion<B>> func) =>
-			a.Optional()
-				.Match(
-					(aVal) => func(aVal),
-					() => new Union<B>());
+		public static Union<T1> Bind<A, T1>(this A a, Func<A, Union<T1>> func) => func(a);
 
-		public static Option<B> Select<A, B>(this IUnion<A> a, Func<A, B> select) =>
-			a.Bind(aval => new Union<B>(select(aval))).opt;
+		public static Union<T1, T2> Bind<A, T1, T2>(this A a, Func<A, Union<T1, T2>> func) => func(a);
 
-		public static IUnion<TSource> Where<TSource>(this IUnion<TSource> source, Func<TSource, bool> predicate) =>
-			source.Optional().Match(
-					(s) => predicate(s) ? source : new Union<TSource>(),
-					() => new Union<TSource>());
+		public static Union<T1, T2, T3> Bind<A, T1, T2, T3>(this A a, Func<A, Union<T1, T2, T3>> func) => func(a);
 
-		public static Option<T> Optional<T>(this IUnion<T> source) =>
-			source == null
-				? new Option<T>()
-				: source.opt;
+		public static Union<T1, T2, T3, T4> Bind<A, T1, T2, T3, T4>(this A a, Func<A, Union<T1, T2, T3, T4>> func) => func(a);
+
+		public static Union<T1, T2, T3, T4, T5> Bind<A, T1, T2, T3, T4, T5>(this A a, Func<A, Union<T1, T2, T3, T4, T5>> func) => func(a);
+
+		public static Union<T1, T2, T3, T4, T5, T6> Bind<A, T1, T2, T3, T4, T5, T6>(this A a, Func<A, Union<T1, T2, T3, T4, T5, T6>> func) => func(a);
+
+		public static Union<T1, T2, T3, T4, T5, T6, T7> Bind<A, T1, T2, T3, T4, T5, T6, T7>(this A a, Func<A, Union<T1, T2, T3, T4, T5, T6, T7>> func) => func(a);
+
+		public static Union<T1, T2, T3, T4, T5, T6, T7, T8> Bind<A, T1, T2, T3, T4, T5, T6, T7, T8>(this A a, Func<A, Union<T1, T2, T3, T4, T5, T6, T7, T8>> func) => func(a);
+
+		public static Union<B> Select<A, B>(this Union<A> source, Func<Union<A>, B> predicate) =>
+			source.Bind(aval => new Union<B>(predicate(aval)));
+
+		//public static IUnion<TSource> Where<TSource>(this IUnion<TSource> source, Func<TSource, bool> predicate) =>
+		//	source.Match(
+		//			(s) => predicate(s) ? source : new Union<TSource>(),
+		//			() => new Union<TSource>());
+
+		//public static Option<T> Optional<T>(this Union source) =>
+		//	source == null
+		//		? new Option<T>()
+		//		: source.opt;
 
 		public static TReturn Match<T1, TReturn>(this Union<T1> union, Func<T1, TReturn> f1)
 		{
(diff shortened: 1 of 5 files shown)


LanguageExt.Core/Prelude_Union.cs
```diff
@@ -0,0 +1,366 @@
+﻿namespace LanguageExt
+{
+	using System;
+
+	public static partial class Prelude
+	{
+		public static IUnion<B> Bind<A, B>(this IUnion<A> a, Func<A, IUnion<B>> func) =>
```
IDisposable: Could we be more consistent with Where here? Seems that `a` and `func` are really `source` and `predicate` like line 16.

sdedalus: I'll clean that up.  I've adapted this from a personal project that was more or less just to see if I could so I'm actually really happy to fix any issues you spot.   I'll fix that naming issue and I've got some more unit tests to add as well.  this will be in my next check-in.

bradphelan:
Love the idea of this extension. However I think "Union" is the wrong name for the type. It is a "discriminated union" not a "union". However that is a sucky long type name. Possibly better is

    OneOf<Cat,Dog,Pencil>
    
rather than

    Union<Cat,Dog,Pencil>

louthy:
I keep meaning to reply to this, but permanently short of time atm!

I don't think this is a strong enough solution for sum-types in C#. I've
tried a number of times in the past to create something with a low
cognitive load, and it's tough. This example suffers from:

* Having to specify the types, in-order, on every usage (which is annoying
at the time, but also a massive pain when refactoring)
* Lack of default handler variants (which if provided cause an exponential
growth in Match overloads)
* As already discussed above, the inheritance approach has holes in that
make the resulting type not as safe as a sum-type in other languages

The closest I've seen to a good implementation is here:
https://functionalcsharp.codeplex.com

It uses the definition of an interface to describe the sum-type. Then uses
reflection emit to dynamically create the sum-type implementation. It still
has its problems, but if I were to add sum-types to Lang-ext then I'd
probably take this approach.

I appreciate you taking the time to contribute to the project, but I don't
think this is strong enough for the library. If you feel like taking the
interface approach I'd happily consider for approval.

P

sdedalus:
I've removed the inheritance problem in my last check-in.  As it is now static checking will throw a compile time error if you have the types out of order or fail to match all cases so the refactoring burden is limited and the chance of a runtime error is minimised.  I'll read over the link you provided It looks like it a solution for tagging the contained type which is nice.  I was working on a wrapper class with proxied operators to manage that aspect of it.

LanguageExt.Core/Prelude_Union.cs
```diff
@@ -0,0 +1,373 @@
...
+		public static Union<T1, T2, T3, T4, T5, T6> Bind<A, T1, T2, T3, T4, T5, T6>(this A a, Func<A, Union<T1, T2, T3, T4, T5, T6>> func) => func(a);
+
+		public static Union<T1, T2, T3, T4, T5, T6, T7> Bind<A, T1, T2, T3, T4, T5, T6, T7>(this A a, Func<A, Union<T1, T2, T3, T4, T5, T6, T7>> func) => func(a);
+
+		public static Union<T1, T2, T3, T4, T5, T6, T7, T8> Bind<A, T1, T2, T3, T4, T5, T6, T7, T8>(this A a, Func<A, Union<T1, T2, T3, T4, T5, T6, T7, T8>> func) => func(a);
+
+		public static Union<B> Select<A, B>(this Union<A> source, Func<Union<A>, B> predicate) =>
+			source.Bind(aval => new Union<B>(predicate(aval)));
+
+		//public static IUnion<TSource> Where<TSource>(this IUnion<TSource> source, Func<TSource, bool> predicate) =>
```
IDisposable: If we don't know how to do this now, record the hoped-for interface elsewhere, but commented-out code is bad.

sdedalus:
IDisposable Louthy seems not very keen on this entire approach so that checkin was intended to capture the final state of my work in case someone finds it useful.   I'm studying the preferred approach now.

louthy:
@sdedalus How do you want to manage this going forward?  Should I close this and you do a new pull request when you're ready, and we can discuss on the Issue that's open?  Or keep this pull request open, and you'll update it as you go?

I think it's worth coming back quickly when you have ideas on a potential approach.  I don't want you wasting time going off in a random direction if it's not something I think will work for the library.  

Elegance is the key to a feature like this working.  At the moment it's possible for users to do this:
```c#
    public abstract class Shape { }

    public class Square : Shape 
    { 
        public readonly Position TopLeft;
        public readonly double Width;
    }

    public class Triangle : Shape 
    { 
        public readonly Position Vertex1;
        public readonly Position Vertex2;
        public readonly Position Vertex3;
    }

    public class Circle : Shape 
    { 
        public readonly Position Position;
        public readonly double Radius;
    }

    var res = x is Square    ? DrawSquare(x as Square)
            : x is Triangle  ? DrawTriangle(x as Triangle)
            : x is Circle    ? DrawCircle(x as Circle)
            : DefaultBehaviour(x);
```

Obviously this isn't such a great example (it's one where inheritance actually works better); but as an example of how to do sum-type like behaviour for C# it's quite elegant.  It has the initial hit of having to build the types, but that's a once only hit, whereas as the initial proposed solution inflicted the declaration on the programmer every time they used it.

It also forces a form of completeness checking, because it's a sequence of ternary operators and therefore you must always have a catch-all case.

And finally it allows for predicate like behaviour:
```c#
    var res = x is Square && (x as Square).Width > 50 ? DrawBigSquare(x as Square)
            : x is Square                             ? DrawSmallSquare(x as Square)
            : x is Triangle                           ? DrawTriangle(x as Triangle)
            : x is Circle                             ? DrawCircle(x as Circle)
            : DefaultBehaviour(x);
```
Obviously it doesn't allow for ad-hoc types like `Union<string, int, bool>`, but I'd argue that you don't get ad-hoc types in F# (or similar functional first languages).  The original `Union` proposal was closer to an `Either` than a sum-type.  

For example in F# you'd be writing this:

```f#
    type Thing = 
      | Integer of int 
      | String  of string
      | Boolean of bool

    let x = Integer 123
    let y = String "Hello"
    let z = Boolean true

    match x with
     | Integer a -> UseInteger a
     | String  a -> UseString a
     | Boolean a -> UseBoolean a
```
Which is the equivalent of:
```c#
    public abstract class Thing { }

    public class Integer : Thing
    {
        public readonly int Value;
        public Integer(int value) { Value = value; }
    }

    public class String : Thing
    {
        public readonly string Value;
        public String(string value) { Value = value; }
    }

    public class Boolean: Thing
    {
        public readonly bool Value;
        public Boolean(int value) { Value = value; }
    }

    var x = new Integer(123);
    var y = new String("Hello");
    var z = new Boolean(true);

    var res = x is Integer  ? UseInteger(x as Integer)
            : x is String   ? UseString(x as String)
            : x is Boolean  ? UseBoolean(x as Boolean)
            : UnknownValue(x);
```
With pattern-matching about to land in C# a lot of this is moot.  The key difficulty is creating the sum-type in the first place.  That's where the interface approach looks quite nice.  Because the above could be declared like so:
```c#
    public interface Thing
    {
        public Thing Integer(int x);
        public Thing String(string x);
        public Thing Boolean(bool x);
    }
```
There is some messiness to the solution on codeplex though:
```c#
    var x = DataType.New<Thing>();
```
This generates an object of type `Thing`.  It would be the equivalent of `null` because it's not until calling one of the methods that you get a concrete value:
```c#
    var x = DataType.New<Thing>().Integer(123);
    var y = DataType.New<Thing>().String("Hello");
    var z = DataType.New<Thing>().Boolean(true);
```
Which is a little unsatisfactory, as this library - if anything - is about correctness and reducing the possibilities of coder fuck-ups.  

The matching story on the codeplex version is quite ugly compared to the examples I gave above using the ternary operators.  

However it does unpack the data, which is quite nice.  That would be good for the `Shape` example:
```c#
    public interface Shape
    {
        Shape Square(Position topLeft, double width);
        Shape Triangle(Position vertex1, Position vertex2, Position vertex3);
        Shape Circle(Position position, double radius);
    }

    // Assuming we had a Prelude function called datatype that does the same as DataType.New
    var x = datatype<Shape>().Square(pos, width);
    var y = datatype<Shape>().Triangle(p1, p2, p3);
    var z = datatype<Shape>().Circle(pos, radius);
    
    var area = x.Match()
        .With(_ => _.Square,   (Position _, double width) =>              AreaOfSquare(width))
        .With(_ => _.Triangle, (Position p1, Position p2, Position p3) => AreaOfTriangle(p1,p2,p3))
        .With(_ => _.Circle,   (Position _, double radius) =>             AreaOfCircle(radius))
        .Return<double>();
```
It's still hella-ugly compared to the ternary approach, and that's without any predicate functionality!  

Having spent the past 20 minutes writing this out (and trying to think of better solutions), I'm definitely erring more towards the position that this is a pointless exercise.  It saves effort on generating the initial type, but the cognitive load is heavy and the existing tools for matching (the ternary operator) is still more powerful and attractive to look at (IMHO).  It is also much more efficient.

The main issue with the ternary approach is having to manually downcast the type, that will be handled with C# 7 pattern-matching which is due very soon.  So I think it's going to be a tough sell unless you can find some magic way to de-ugly all this, support predicates, have completeness checking, _and_ be efficient.  

Your call, but I think it'll be tough!

sdedalus:
Go ahead and close this. I have some ideas that need a bit of fleshing out
before they are ready and honestly I'm having fun putting the type system
through it's paces so I'll keep working on it regardless.  If and when I
get this to a state where it seems like it could benefit the project I'll
run it by you on the issue board assuming that it hasn't been closed also.

louthy:
> and honestly I'm having fun putting the type system through it's paces so I'll keep working on it regardless

Yeah, it's always fun pushing the language as far as it can go (it's where this library came from!).  Let me know how you get on.

sdedalus:
Would a syntax like the following meet the criteria mentioned above?

```
Union<string, int> x = 100;
string value = x.Match<string>()
	.Where(v => v)   // This only matches a string type and returns that string
	.Where(c => c == 100, v => "Keeping It 100.") // when the value is 100 this returns.
	.Where(v => "Tea?") // if the previous value isn't a match this.
	.Else(() => "Nothing"); // if nothing else matches this.

Assert.Equal("Keeping It 100.", value); 
```

This enforces that all possible states are covered.  When you provide a matching condition the interface won't provide a way to match the next type or retrieve the output value until you provide a default for that type and the final value is not retrievable unless you provide a default handler if nothing else matches. 

the following is also possible:

var x = new Union<string, Union<string, Union<string>>>(
	new Union<string, Union<string>>(
		new Union<string>("Test")));

string value = x.Match<string>()
	.With(v => "Not This")
	.With(v => v
.		Match<string>()
		.With(b => "Not This")
			.With(c => c.Match<string>()
 			.With(d => d) // this is the match
			.Else(() => "Not This"))
		.Else(() => "Not This"))
	.Else(() => "Not This");

	Assert.Equal("Test", value);

PR #162, opened 2016-10-17
Add composition overloads

### LINKED ISSUE #125: Implement composition for 0 arity and void returning functions
Currently, the function composition extension methods cannot be invoked against zero arity functions or void returning functions:

```
var readline = fun(Console.ReadLine);
var readlineDummyArg = fun((Unit _) => Console.ReadLine());

readline.BackCompose(File.ReadAllBytes); // Cannot resolve matching overload
readlineDummyArg.BackCompose(File.ReadAllBytes);
```

A full set of extension methods would look like the following:

```
public static class ___ComposeExt
{
    public static Func<TOut> Compose<TMid, TOut>(this Func<TMid> f, Func<TMid, TOut> g) 
        => () => g(f());
    public static Func<TIn, TOut> Compose<TIn, TMid, TOut>(this Func<TIn, TMid> f, Func<TMid, TOut> g) 
        => arg => g(f(arg));

    public static Action Compose<TMid>(this Func<TMid> f, Action<TMid> g) 
        => () => g(f());
    public static Action<TIn> Compose<TIn, TMid>(this Func<TIn, TMid> f, Action<TMid> g) 
        => arg => g(f(arg));
}
```

Since one of the goals of the library is to eliminate `void`, the latter two can be substituted with:

```
    public static class ___ComposeExt
    {
   ...
        public static Func<Unit> Compose<TMid>(this Func<TMid> f, Action<TMid> g)
            => () => { g(f()); return Unit.Default; };
        public static Func<TIn, Unit> Compose<TIn, TMid>(this Func<TIn, TMid> f, Action<TMid> g)
            => arg => { g(f(arg)); return Unit.Default; };
    }
```

I realize my terminology in the snippets is a bit wrong because what I term `Compose` is actually `BackCompose`, but you get the gist. In total we should end up with 4 overloads for `Compose` and 4 overloads for `BackCompose`, plus possibly additional overloads for the `compose` function introduced by `Prelude`.

masaeedu:
Add overloads for zero-arity and void returning functions.

Fixes #125

### DIFF
--- a/LanguageExt.Core/Compose.cs
+++ b/LanguageExt.Core/Compose.cs
@@ -26,7 +26,7 @@ namespace LanguageExt
             () => b(a());
 
         /// <summary>
-        /// Function composition
+        /// Action composition
         /// </summary>
         /// <returns>v => b(a(v))</returns>
         [Pure]
@@ -34,7 +34,7 @@ namespace LanguageExt
             => v => { b(a(v)); return Unit.Default; };
 
         /// <summary>
-        /// Function composition
+        /// Action composition
         /// </summary>
         /// <returns>() => b(a())</returns>
         [Pure]
@@ -44,7 +44,7 @@ namespace LanguageExt
         /// <summary>
         /// Function back composition
         /// </summary>
-        /// <returns>b(a(v))</returns>
+        /// <returns>v => b(a(v))</returns>
         [Pure]
         public static Func<T1, T3> BackCompose<T1, T2, T3>(this Func<T1, T2> a, Func<T2, T3> b) =>
             v => b(a(v));
@@ -52,13 +52,13 @@ namespace LanguageExt
         /// <summary>
         /// Function back composition
         /// </summary>
-        /// <returns>b(a())</returns>
+        /// <returns>() => b(a())</returns>
         [Pure]
         public static Func<T2> BackCompose<T1, T2>(this Func<T1> a, Func<T1, T2> b) =>
             () => b(a());
 
         /// <summary>
-        /// Function back composition
+        /// Action back composition
         /// </summary>
         /// <returns>v => b(a(v))</returns>
         [Pure]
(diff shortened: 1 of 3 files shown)


IDisposable:
Would love to see some unit-test coverage for these.

louthy:
@masaeedu Hi Asad, could you please provide unit tests for your new functionality.  Thanks,

masaeedu:
@louthy @IDisposable Yeah, working on it. There wasn't any unit tests for the existing `Compose`/`BackCompose` extension methods nor the `Prelude.compose` static overloads; I'll add coverage for the `Compose` and `BackCompose` extension methods, but you probably want to use codegen for methods with lots of overloads like `Prelude.compose`.

masaeedu:
@louthy Done. The CI output looks kind of weird, it's spewing lots of HTML, but I ran the tests locally and they compile & pass.

louthy:
@masaeedu The CI is broken atm, haven't gotten around to fixing it.  I'll grab locally and test.

LanguageExt.Core/Compose.cs
```diff
@@ -12,17 +12,65 @@ public static class __ComposeExt
...
+        [Pure]
+        public static Func<T2> Compose<T1, T2>(this Func<T1, T2> b, Func<T1> a) =>
+            () => b(a());
+
+        /// <summary>
+        /// Action composition
+        /// </summary>
+        /// <returns>v => b(a(v))</returns>
+        [Pure]
+        public static Func<T1, Unit> Compose<T1, T2>(this Action<T2> b, Func<T1, T2> a)
```
masaeedu: I'm fine with having an `Action` return value, I was just trying to be in line with the library's goal of eliminating void. Note that we can't have it both ways; two overloads can't differ only in return type. Keeping that in mind, should we go ahead and change:

```
public static Func<T1, Unit> Compose<T1, T2>(this Action<T2> b, Func<T1, T2> a)
```

to:

```
public static Action<T1> Compose<T1, T2>(this Action<T2> b, Func<T1, T2> a)
```

?

masaeedu: Also, you don't need a separate overload for an argument of type `this Func<T2, Unit>`, that's already covered by the existing single argument function composition overloads.

masaeedu:
@louthy ping

louthy:
@masaeedu 

Looks good, thanks Asad

Issue #164, opened 2016-10-20
Make unit tests run against .NET Core


masaeedu:
`LanguageExt.Tests` is currently only built and run against .NET 4.5. Since most parts of this library are now building against .NET Standard (the missing pieces are tracked in #70), it should also be unit tested on .NET Core. There are currently a few minor incompatibilities in the `LanguageExt.Tests` project that prevent it being run against `netcoreapp1.0` in addition to `net45`:
- Uses `WebClient`. This can be swapped out for the .NET Standard compatible `System.Net.Http.HttpClient`
- Uses `SystemException` as an arbitrary exception in some exception related unit tests. This can be swapped for any other .NET Standard compatible exception (e.g. `ArgumentException`)
- Solution has redundant project files which cause `dotnet restore` for the unit test project to fail due to ambiguity
- Uses overload signatures that are only defined behind an `#if !COREFX` directive (see for example [this overload](https://github.com/louthy/language-ext/blob/7be100c/LanguageExt.Process/Prelude_Spawn.cs#L303) of `spawn`, which is used [here](https://github.com/louthy/language-ext/blob/7be100c/LanguageExt.Tests/ProcessProxies.cs#L68)). These can either be fixed by providing equivalent overloads under an `#else` directive, or by putting the entire unit test under a `#if !COREFX` directive

I've fixed most of these and will set up a PR for review.

PR #165, opened 2016-10-20
Remove redundant project files

masaeedu:
- Duplicate project.json files are removed
- All csproj, fsproj, xproj, and sln files are removed. xproj and sln files can be automatically generated by Visual Studio and do not need to be in source control, and csproj/fsproj files are redundant.

Part of fix for #164.


louthy:
I'm sorry I can't accept this right now.  I am making major changes in another branch and this will cause conflicts that I really don't want to be resolving.  I appreciate the effort, but it's probably best if you shout on the issues page before ploughing ahead with changes.

masaeedu:
@louthy I opened an issue in #164, but fair enough. Usually if there are conflicts you can ask the person making the PR to rebase once you've merged in whatever you're prioritizing.

masaeedu:
@louthy Btw, one of the reasons for making this change is so all files aren't listed in a csproj/fsproj file, causing merge headaches. This allows project structure changes and code changes to be made orthogonally. Unless you're moving/adding new projects or modifying dependencies, merging in this series of changes should not cause conflicts. Is the branch you're working on https://github.com/louthy/language-ext/tree/type-classes?

louthy:
> Usually if there are conflicts you can ask the person making the PR to rebase once you've merged in whatever you're prioritizing.

You don't understand the scope of the changes being made - it's borderline a re-write.  I really don't want to have to deal with the merge issues.

masaeedu:
@louthy Assuming you're talking about the `type-classes` branch, I've gone over all the changes and they're perfectly mergeable. I don't think you're quite grokking what the change is doing. Nearly all of your changes in the csproj files are *modifying the list of files to be compiled*. This list no longer needs to exist; you can just delete the csproj file and continue working. The project.json format does not need to maintain a list of all files in the project.

If you don't want to think about this right now, would you instead be willing to take a mergeable PR to the type-classes branch?

louthy:
This is something I really don't want to be thinking about at the moment.  My brain is not in this space as I've got a 1000 plates spinning with the new version.  Just a quick glance at your changes and they appear to be very dangerous, deleting all the csproj files, which is insane (The non-Core sln exists because compiling the .NET Core solution is so slow).  So please leave it for now, I'm not quite sure what your intentions are and I'm not able to dedicate enough time to check.

masaeedu:
> Just a quick glance at your changes and they appear to be very dangerous, deleting all the csproj files, which is insane 

The project.json format was intended as a replacement for csproj, so I don't know where you're getting that deleting csproj is "insane". Having both project.json and csproj is redundant, and it leads to configuration drift (this has already started happening with your dependencies). After applying the changes I made to master, the solution compiles in both Visual Studio and from the command line, and all unit tests not requiring a redis instance pass against `netcoreapp1.0` and `net4.5` targets. 

Don't really know what else to tell you here, maybe google the docs on project.json. A few months from now you could be stuck consolidating your project.json and csproj files and migrating to the new msbuild format by hand, which is going to be a pain in the ass.

> (The non-Core sln exists because compiling the .NET Core solution is so slow)

Yes, this is a problem lots of people have complained about. They're tracking it in dotnet/cli#2515, but this is only going to improve after they switch back to msbuild in 2017.

> This is something I really don't want to be thinking about at the moment. My brain is not in this space as I've got a 1000 plates spinning with the new version.

Hey, its your library, do as you will.

louthy:
> Having both project.json and csproj is redundant

No it isn't.  As the person who spends the most time in the project compiling and testing, I can tell you that the csproj is vital for rapid iteration and testing.  The project.json is the deployment project, and it works great for that because it supports the myriad .NET versions and makes building of nuget packages easy -- but it's slow.

>  it leads to configuration drift (this has already started happening with your dependencies)

Doesn't matter, as the deployed version comes from one SLN.  The only dependencies are slow moving ones (like Newtonsoft.JSON, Rx.NET, etc), the APIs are static.  So even if they fall behind a bit on the csproj solution, it's fine.  I run the unit-tests before building the nuget packages.  So the project.json can be considered the 'master' solution (even if I spend most of my time in the csproj based solution for coding convenience).

> A few months from now you could be stuck consolidating your project.json and csproj 

Another reason I don't want any movement on this until MS sort out their project system.  The settings in the project.jsons were guided by P Carter from the F# team at MS - I'm happy to keep them as-is until MS sort their tooling out. I wasted so much time with DNX and then .NETCore, I am _never_ making the mistake of pre-empting them again.  I'll wait until v1 of the tools are actually ready (and will probably still wait a while after that).

> Yes, this is a problem lots of people have complained about. 

So you knew about it, and decided to cripple my build process anyway?  Thanks!  I don't want to sound ungrateful, any help on the project is welcome, but it's pretty reckless to come into a mature project and just change the entire project setup without any consideration or consultation of the people who have to maintain and deploy it.  It's a big jump from adding a few composition functions to changing my entire workflow.

masaeedu:
> So you knew about it, and decided to cripple my build process anyway? Thanks

Jesus. On my laptop, it takes ~50 seconds to build `LanguageExt.Core` (which is the bottleneck) for a single target using `dotnet build`. It takes ~30 seconds to build it using `msbuild`. This doesn't seem to me like an earth-shattering performance regression, but YMMV. Typical build times for the projects I work on are 20-30 seconds, so in general I have used `project.json` wherever possible.

But you caught me. My real intention was to "insanely" "cripple" your build process, as I twirled my moustache and cackled maniacally.

PR #171, opened 2016-10-31
Fix for == methods in NewType and Pos throwing NullReferenceExceptions

Richiban:
The types NewType<> and Pos both throw exceptions if the == operator is used and the left-hand side is null. E.g.

```csharp

public class Name: NewType<string> { public Name(string value) : base (value){} }

var x = (Name) null;

if ( x == null ) // Throws NullReferenceException
{ }

```

### DIFF
--- a/LanguageExt.Core/NewType.cs
+++ b/LanguageExt.Core/NewType.cs
@@ -57,10 +57,8 @@ namespace LanguageExt
             Value.Equals(other.Value);
 
         [Pure]
-        public override bool Equals(object obj) =>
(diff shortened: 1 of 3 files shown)


louthy:
@Richiban Hi Richard, thanks for this.  But I'm going to reject it.  The code is fine, it's just this is already fixed on the type-classes branch, which will be the next release:

https://github.com/louthy/language-ext/blob/type-classes/LanguageExt.Core/DataTypes/NewType/NewType.Predicate.cs#L68

Richiban:
Sure, no problem!

PR #178, opened 2016-12-04
Trampoline recursion

ncthbrt:
This is a simple addition to the core library. It allows users to define tail recursive functions which can recurse to a much greater depth than that allowed naturally by C#.
For example, a user would define a function, such as:

```cs
public static Recursive<Map<string,int>> Elections(Map<string,long> votes, IEnumerable<string> candidates) => 
  candidates.Any()
  ? Recurse(Elections, CountVotes(votes, candidates.Head()), candates.Tail())
  : Return(votes);
```
When a user unwraps the value through a direct cast or implicit conversion, the function is evaluated using [trampolining](https://clojuredocs.org/clojure.core/trampoline). The Recurse and Return helper functions make trampolining transparent to the end user.

### DIFF
--- a/LanguageExt.Core/DataTypes/Recursive/Recursive.cs
+++ b/LanguageExt.Core/DataTypes/Recursive/Recursive.cs
@@ -15,9 +15,7 @@ namespace LanguageExt.DataTypes.Recursive
     /// <remarks>
     /// Intended usage is for situations where an algorithm might more conveniently be expressed as 
     /// a recursive function but fears this may result in a stack overflow error.
-    /// 
-    /// Sample 
-    /// 
+    ///     
     /// Note: This is a lazy type. The function is only invoked when an implicit cast occurs    
     /// </remarks>
     /// <typeparam name="A">Base case return type</typeparam>


ncthbrt:
Note: Failing build appears unrelated to additions on this branch

LanguageExt.Core/DataTypes/Recursive/Recursive.Prelude.cs
```diff
@@ -0,0 +1,451 @@
...
+    public static partial class Prelude
+    {
+
+        /// <summary>
+        /// Returns the base of the recursion
+        /// </summary>
+        /// <typeparam name="A"></typeparam>
+        /// <param name="t"></param>
+        /// <returns></returns>
+        public static Recursive<A> Return<A>(A a) => new Recursive<A>(a);
```
louthy: Is `Return` the best name for the constructor function here?  `Return` is known for being a monad constructor function, so I think this could be confusing, and will probably clash with [the new monad type-class](https://github.com/louthy/language-ext/blob/type-classes/LanguageExt.Core/TypeClasses/Monad/Monad.Prelude.cs#L20).  Perhaps `Recurse` because, regardless of its parameter, it's returning a `Recursive<A>`?

LanguageExt.Core/DataTypes/Recursive/Recursive.cs
```diff
@@ -0,0 +1,68 @@
...
+        {
+            Recurse,
+            Return
+        }
+
+        private readonly State recursiveState;
+        private readonly Func<Recursive<A>> recursiveFunction;
+        private readonly A val;
+
+        internal Recursive(Some<A> val)
```
louthy: Please remove `Some<A>`, it isn't used throughout language-ext due to its limitations.  Please just check the `val` for `null` and `throw`

LanguageExt.Core/DataTypes/Recursive/Recursive.cs
```diff
@@ -0,0 +1,68 @@
...
+        {
+            this.val = val;
+            recursiveFunction = default(Func<Recursive<A>>);
+            recursiveState = State.Return;
+        }
+
+        internal Recursive(Func<Recursive<A>> func)
```
louthy: Check `func` for `null`

LanguageExt.Core/DataTypes/Recursive/Recursive.cs
```diff
@@ -0,0 +1,68 @@
...
+            recursiveFunction = func;
+            val = default(A);
+        }
+
+        public static implicit operator A(Recursive<A> val)
+        {
+            // Here is where the trampolining (and execution takes place).
+            // The `current` function is invoked which yields either a result of type 'A or the next function to invoke.
+            
+            Some<Recursive<A>> current = val;
```
louthy: Remove use of `Some<A>`, it adds a (admittedly small) unnecessary overhead to the execution.

LanguageExt.Core/DataTypes/Recursive/Recursive.cs
```diff
@@ -0,0 +1,68 @@
...
+            while (current.Value.recursiveState != State.Return)
+            {
+                current = current.Value.recursiveFunction();
+            }
+            return current.Value.val;
+        }
+
+        [Pure]
+        public static implicit operator Recursive<A>(A val) => new Recursive<A>(val);
```
louthy: Please place the body of the member expression on a new line:
```c#
    public static implicit operator Recursive<A>(A val) => 
        new Recursive<A>(val);
```

LanguageExt.Core/DataTypes/Recursive/Recursive.cs
```diff
@@ -0,0 +1,68 @@
...
+                current = current.Value.recursiveFunction();
+            }
+            return current.Value.val;
+        }
+
+        [Pure]
+        public static implicit operator Recursive<A>(A val) => new Recursive<A>(val);        
+
+        [Pure]
+        public Type GetUnderlyingType() => typeof(A);
```
louthy: Remove this.  It's available in `Option` to support `MatchUntyped`, this behaviour isn't needed here.

LanguageExt.Core/DataTypes/Recursive/Recursive.cs
```diff
@@ -0,0 +1,68 @@
...
+    /// This class is implemented using trampoling and implicit type conversion.   
+    /// </summary>    
+    /// <remarks>
+    /// Intended usage is for situations where an algorithm might more conveniently be expressed as 
+    /// a recursive function but fears this may result in a stack overflow error.
+    ///     
+    /// Note: This is a lazy type. The function is only invoked when an implicit cast occurs    
+    /// </remarks>
+    /// <typeparam name="A">Base case return type</typeparam>
+    public class Recursive<A>
```
louthy: Can we rename this `Rec<A>`?

LanguageExt.Core/DataTypes/Recursive/Recursive.Prelude.cs
```diff
@@ -0,0 +1,451 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using LanguageExt.DataTypes.Recursive;
+
+namespace LanguageExt
+{
+    public static partial class Prelude
```
louthy: In general the comments for `Return` and `Recurse` are either missing or very light.  They're constructor functions, and therefore should have some more info about their purpose (like the `Recursive<A>` type).

LanguageExt.Core/DataTypes/Recursive/Recursive.cs
ncthbrt: Will do

ncthbrt:
I Will make the requested changes in the next few days. Got bogged in some work as well

louthy:
@ncthbrt No worries, there's no rush.  It'll be a little while until I get this branch released.  Make sure you grab the latest because I have fixed up a load of references on the csproj build.

LanguageExt.Core/DataTypes/Recursive/Recursive.Prelude.cs
ncthbrt: I wanted to make it clear that this is the base of the recursion. But I'm happy to change it if you feel it is more consistent

louthy: I open to suggestions, I just think that `Return` isn't really the right name here.  I think whatever it is should be a single word, so no `RecReturn`, `RecurseReturn` type ugliness.  We should also avoid anything too generic like `Result`.  Thoughts?

IDisposable: `Yield` is one choice, but might confer the wrong idea. Perhaps `Produce`?

louthy: @IDisposable I quite like `Yield`

ncthbrt: I quite like `Yield`, but it might be confused with the `yield` keyword.

louthy: Also `Yield` is a verb rather than noun, so for a constructor name it's not ideal.  And with that in mind `Recurse` is a bad choice also.  We need something like `Some`, `None`, `Left`, `Right`, something that gives an indication of the state of the `Recursive` once constructed.

The only things that come to mind are 'termination' synonyms: `Final`, `End`, `Last`, `Done`, etc.

IDisposable: Technically, `Yield` is also a noun (says the guy doing tons of analysis on corn yields). For a less _dangerous_ noun, perhaps `Outcome`, `Issue`, `Product` or `Harvest` (okay, maybe a bit corny)

louthy: @IDisposable 

> Technically, Yield is also a noun

That's a good point... I think I'm back on-board with `Yield`! (I'm so fickle)

ncthbrt:
Hey. Sorry it is taking such a while. Still planning to make the changes. Just got some pretty crushing deadlines at work at the moment.

louthy:
@ncthbrt No worries, take your time.

shmutalov:
Ping

Issue #183, opened 2016-12-26
Haskell flip function analogue


leandromoh:
I saw this project has methods that allow the programmers to use functions as curried and partial applications.

So I thought that would be also cool has a Flip function (like the Haskell one) that works for both curried and uncurried functions.

I'm planning to submit a PR for this.

What do you think?

louthy:
As with uncurry, it's likely to have limited use, but again it's low maintenance, so yes go ahead.  Again, could you make the changes in the type-classes branch.

louthy:
I made some changes to [your pull request](https://github.com/louthy/language-ext/pull/184).  The latest is now in `v2.0.31-beta`.

PR #184, opened 2016-12-31
implemented Flip

leandromoh:
resolved issue #183

### DIFF
--- a/LanguageExt.Core/Prelude_Flip.cs
+++ b/LanguageExt.Core/Prelude_Flip.cs
@@ -1,0 +1,152 @@
+using System;
+using System.Diagnostics.Contracts;
+
+namespace LanguageExt
+{
+    public static partial class Prelude
+    {
+        /// <summary>
+        /// Reverse the order of the first two arguments of a curried function
+        /// </summary>
+        [Pure]
+        public static Func<T2, Func<T1, TResult>> flip<T1, T2, TResult>(this Func<T1, Func<T2, TResult>> function)
+        {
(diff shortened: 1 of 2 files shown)


louthy:
@leandromoh Hi Leandro, just to let you know that I haven't forgotten about your pull requests.  I have a ton of work to do to get the project stable again after moving to VS2017.  Once completed, I'll be reviewing your reqs.

leandromoh:
@louthy  no problem. thanks for the explanation

louthy:
@leandromoh  Hi Leandro, I have modified this.  In Haskell `flip` works on two arguments only.  There is a clear reason for that which is composition and application.  It's less important in C# as I mentioned in the original issue.  However I see a value in flipping the first and the _last_ argument.  Especially if C# gets the forward pipe operator that is currently [a proposal on Rosyln](https://github.com/dotnet/roslyn/issues/5445).  However beyond 3 arguments I think flipping the first and the last argument feels wrong.  So I have modified this to only work on:
```c#
    Func<A, Func<B, R>>           ---> Func<B, Func<A, R>>
    Func<A, Func<B, Func<C, R>>>  ---> Func<C, Func<B, Func<A, R>>>

    Func<A, B, R>                 ---> Func<B, A, R>
    Func<A, B, C, R>              ---> Func<C, B, A, R>
```
There's a nice symmetry with 2 or 3 arguments which doesn't exist with more.  

[You can find it here](https://github.com/louthy/language-ext/blob/master/LanguageExt.Core/Prelude/Prelude_Flip.cs)

PR #192, opened 2017-02-03
window.location.origin is (sometimes) undefined

StanJav:
`window.location.origin` isn't supported by all browsers

See: 
https://developer.mozilla.org/en-US/docs/Web/API/HTMLHyperlinkElementUtils/origin
https://connect.microsoft.com/IE/feedback/details/1763802/location-origin-is-undefined-in-ie-11-on-windows-10-but-works-on-windows-7


louthy:
I've have manually patched this on echo-process

PR #211, opened 2017-04-22
Added parseEnum

jeremysmith1:
This is my attempt at parseEnum function. There is a constraint on Enum that is must be a non-nullable value type. I added the struct constraint so that would not be an issue. The downside being that you have to use it like so, passing in the enum type
`enum breakfast { Eggs, Bacon, Toast };` `var test = parseEnum<breakfast>("Toast");`


louthy:
@jeremysmith1 Thanks Jeremy!

PR #239, opened 2017-07-10
Serialization fix for Option<A>

StanJav:
The previous version thrown ValueIsNoneException.


louthy:
This won't work.  But I see the issue you're trying to fix, so I will do it.

PR #248, opened 2017-08-11
Mark generated files as generated

sharwell:
This change prevents several analyzers from running on the code, causing delays to the overall IDE experience which are particularly observable in Visual Studio 2017.


louthy:
@sharwell Thanks Sam.

Issue #249, opened 2017-08-12
Solution performance in Visual Studio (discussion)


sharwell:
Someone pointed me to this repository yesterday as an example of a project which has "interesting performance characteristics" in the IDE. I wanted to open a discussion so I can make some notes during my investigation as well as point out some issues that are likely to benefit this project.

The first two items that substantially improve the overall experience for this project were dotnet/roslyn#19258 (part of the 15.3 update release), followed by #248 here. The former reduces the amount of work performed by analyzers to evaluate dotted expressions, such as `Trans<MArr<Arr<A>>, Arr<Arr<A>>, MArr<A>, Arr<A>, A>.Inst.Bind`. The second disables most analyzers (code style and usage) altogether in code which isn't intended to be edited anyway.

The next area where I believe this project is currently negatively influenced involves documentation comments. While I can make some improvements to this quickly, I believe that certain options regarding comments are limited by the language itself. I would encourage you to express support for dotnet/csharplang#401.

I'll be watching for general places where things can be improved, but overall I suspect it will be many small pieces of an overall big puzzle. If you have any specific IDE performance- or behavior-related concerns related to this project you'd like me to look at, please feel free to add them here.

louthy:
@sharwell Thanks for looking into this.  I do have major performance issues with VS 2017, this project is only one of my problems.  My main work project, which unfortunately I can't share, is constantly locking up and has laggy text input.  

One of my suspicions was expressions in general, but especially readonly statics.  I have a project where I use an in-house virtual dom system that bakes to readonly statics, i.e.
```c#
public readonly static DomMap Header =
    WithAppointment(
        SectionHeader("Appointment", "{AppointmentType}", "{prettydate(TimelineDate)}"),
        Details(
            Info("Date and time",
                If<AppointmentModel>(a => a.ArrivedDateTime.IsSome,
                    Then("{shorttime(Start)} to {shorttime(Finish)} on {shortdate(Start)}, arrived at {shorttime(ArrivedDateTime)} on {shortdate(ArrivedDateTime)}"),
                    Else("{shorttime(Start)} to {shorttime(Finish)} on {shortdate(Start)}, not arrived"))),
            Info("Site and location", "{Site}: {Location | (location not specified)}"),
            Info("Attendees", Attendees),
            Where<AppointmentModel>(am => HasServices(am, AppointmentServiceState.Provided),
                Info("Provided services", Services(AppointmentServiceState.Provided))),
[...]

PR #250, opened 2017-08-12
XML-escape replacement sequences in HKT.Extensions and HKT.Prelude

sharwell:
Fixes one of the major cases of invalid XML contained in documentation comments.

❓ @louthy For changing `` `Num<A>` `` in the templates, I can either change it to `` `Num&lt;A&gt;` `` (follows your doc tool) or to `<see cref="Num{A}"/>` ("proper" XML doc syntax). Let me know your preference and I will add a commit to this pull request.

This is related to the performance investigations discussed in #249.


LanguageExt.Core/HKT/HKT.Extensions.tt
```diff
@@ -142,6 +142,37 @@ namespace LanguageExt
 }
 <#+
 
+    public struct StringWrapper
```
sharwell: :memo: I didn't see a way to define an extension method for use in T4, so I went "sneaky" instead.

louthy:
@sharwell Thanks Sam, I'll give this a go.

PR #256, opened 2017-08-23
Fix Result<T> when created with a null exception

heemskerkerik:
Currently, when you create a `Result<T>` with a null exception, it is considered successful. This PR fixes that.

### DIFF
--- a/LanguageExt.Tests/ResultTests.cs
+++ b/LanguageExt.Tests/ResultTests.cs
@@ -9,35 +9,35 @@ namespace LanguageExt.Tests
         [Fact]
         public void TestBottom()
         {
-            Assert.True(_bottomResult.IsFaulted);
-            Assert.True(_bottomResult.IsBottom);
+            Assert.True(bottomResult.IsFaulted);
+            Assert.True(bottomResult.IsBottom);
         }
 
         [Fact]
         public void TestSuccess()
         {
-            Assert.False(_successResult.IsFaulted);
-            Assert.False(_successResult.IsBottom);
+            Assert.False(successResult.IsFaulted);
+            Assert.False(successResult.IsBottom);
         }
 
         [Fact]
         public void TestFault()
         {
-            Assert.True(_faultResult.IsFaulted);
-            Assert.False(_faultResult.IsBottom);
+            Assert.True(faultResult.IsFaulted);
+            Assert.False(faultResult.IsBottom);
         }
 
         [Fact]
         public void TestFaultWithNullException()
         {
-            Assert.True(_faultWithNullException.IsFaulted);
-            Assert.False(_faultWithNullException.IsBottom);
+            Assert.True(faultWithNullException.IsFaulted);
+            Assert.False(faultWithNullException.IsBottom);
         }
 
         [Fact]
         public void TestMatchWithBottom()
         {
-            string output = _bottomResult.Match(
+            string output = bottomResult.Match(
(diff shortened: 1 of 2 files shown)


LanguageExt.Tests/ResultTests.cs
```diff
@@ -0,0 +1,294 @@
...
+        [Fact]
+        public void TestMapWithFaultWithNullException()
+        {
+            var output = _faultWithNullException.Map(_ => 2);
+            
+            Assert.True(output.IsFaulted);
+            Assert.False(output.IsBottom);
+        }
+
+        private readonly Result<int> _bottomResult = new Result<int>();
```
louthy: Could you remove the leading underscores from the names?  It doesn't fit with any of the existing code base and it's a style I disagree quite strongly with.  Thanks.

heemskerkerik: No problem. Fixed.

louthy:
@heemskerkerik Thanks Erik 👍

LanguageExt.Tests/ResultTests.cs
jeffward01: > Could you remove the leading underscores from the names? It doesn't fit with any of the existing code base and it's a style I disagree quite strongly with. Thanks.

I know I'm reviving an old issue, and the chances of you seeing this are slim to none, but i'm curious about your opinion of underscores:

In the microsoft coding conventions it seems to be 'the proper way to work with private variables' 

> Use camel casing ("camelCasing") when naming private or internal fields, and prefix them with _.

https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/coding-conventions#camel-case 

-----

What do you think?

louthy: The 'proper way' is just an opinion, like what I am about to say is just an opinion.  Microsoft don't have the monopoly on wisdom, and blindly following what MS say isn't necessarily going to work out great for you; look at their history of failures that were pushed as 'the one true way' in the past.  

_Trust yourself and your ability to make good judgements about what works for you._

The underscores in front of private variables are a hangover from [the MFC days](https://en.wikipedia.org/wiki/Microsoft_Foundation_Class_Library) of [Hungarian Notation](https://en.wikipedia.org/wiki/Hungarian_notation).  Hungarian Notation was a way of baking in details of the type of the variable into its name: `szName` for a 'name' field that is a `string`.  It also had the concept of variable origin: `m_szName` is a string field in a class, which was also sometimes done as `_szName`; `g_szName` is global; `s_szName` is static, etc.

Hungarian Notation added nothing but overhead really (overhead for the dev that is).  It also meant tooling like intellisense had to get past several characters before getting to the subject of the search.  

The key point is this: _Is your code so complex that you can't tell what it does without embedding type-names and variable-origins into the name?  If so, that's a reason to make the code simpler, not to start embedding meta-data into variable names._

When .NET came along and the BCL was created for the first time, the days of Hungarian Notation were shelved.  But the underscore persisted.  There's no good reason for it _in my personal opinion_, if it clashes with a local then prefix it with `this.` - that is a much more declarative way of showing where the variable came from.

By the way, I believe even Microsoft dropped the underscore requirement (for their internal code-bases) at one point.  However, I do see it used in projects like Roslyn, so I'm not sure how long that lived or whether it's optional per-project internally.  But it highlights that this is just opinionated devs, not necessarily an immutable truth.  Make your own judgement.

PR #260, opened 2017-08-27
Adds async constructors to TryOptionAsync and TryAsync

NickDarvey:
If you have a function that returns a `Task<>` and try and use the existing constructor functions you'll end up with that `Task<>` wrapped inside the `OptionResult<>` or `Result<>`. This PR adds constructors to allow the result to be correctly wrapped.

Unfortunately, I've had to choose a different name to disambiguate because C#'s type inference cannot tell which constructor should be used. Honestly, I wanted to ditch the `Task.Run(() => {})` constructors because people could explicitly write that if they want it to run on the thread pool---however, that'd be a breaking change so I avoided that.
(If you have better approach for this, please let me know.)

### DIFF
--- a/LanguageExt.Core/DataTypes/TryAsync/TryAsync.Prelude.cs
+++ b/LanguageExt.Core/DataTypes/TryAsync/TryAsync.Prelude.cs
@@ -17,9 +17,21 @@ namespace LanguageExt
         /// <param name="f">Function to run asynchronously</param>
         /// <returns>A lifted operation that returns a value of A</returns>
         [Pure]
-        public static TryAsync<A> TryAsync<A>(Func<A> f) => 
+        public static TryAsync<A> TryAsync<A>(Func<A> f) =>
             TryAsyncExtensions.Memo<A>(() =>
+            
                 Task.Run(() => new Result<A>(f())));
+
+        /// <summary>
+        /// TryAsync constructor function
+        /// </summary>
+        /// <typeparam name="A">Bound value type</typeparam>
+        /// <param name="f">Asynchronous function to run asynchronously</param>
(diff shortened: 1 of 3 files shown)


louthy:
The AsyncAsync naming troubles me.  Let me have a think about this.  

I am in two minds about the Task.Run.  One one hand you wouldn't be making a Func asynchronous unless you wanted it to run asynchronously, so Task.Run makes sense.  On the other hand it could be that in reality its a convenience function for use in LINQ (which is probably what I really added it for - but my memory is letting me down today).

I'm more of the mind to remove the constructor function altogether, because its effects are unclear.  More thought, I'll get back to you.

NickDarvey:
Is there another way I can create a TryOptionAsync when I already have a function that is task-returning (`Func<Task<Option<string>>>`)?

NickDarvey:
Closing because there is this exists: [ToAsync :: Try<Task<T>> -> TryAsync<T>](https://github.com/louthy/language-ext/blob/master/LanguageExt.Core/DataTypes/Try/Task.Try.Extensions.cs#L32).

Spotted by @trbngr in #283.

cfiet:
Hi @NickDarvey, @louthy 

I still feel that those constructors would be very useful, especially if implemented with `ToAsync` calls.

I think `ToAsync` is not the clearest way of doing that, hence this issue as well as #283 . Constructors on `Prelude` were the first place I looked and to be honest, I did create `Try<Task<T>>` as an attempt of solving my issue, but missed the `ToAsync` until it was pointed out to me.

I had an idea of implementing `Try*Task` methods to keep existing `Try*Async`s intact, but since they are already implemented here, I think it would be worth refining those.

PR #282, opened 2017-10-18
Static properties cause  Record<T>::Equals to fail

### LINKED ISSUE #281: static properties cause Record::Equals to crash
I have the following test case

```
using static LanguageExt.Prelude;using LanguageExt;
using LanguageExt;
using Xunit;
using FluentAssertions;
using Ninject.Infrastructure.Language;
using static LanguageExt.Prelude;

namespace Utils.Spec
{

    public class LanguageExtRecords
    {
        public class Foo : Record<Foo>
        {
            public static Foo Default { get; private set; }

            static Foo()
            {
                Default = new Foo( 10,20,List(1,2) );
            }

            public Foo(int age, int s, Lst<int> numbers)
            {
                Age = age;
                String = s;
                Numbers = numbers;
            }
            public int Age { get; }
            public int String { get; }
            public Lst<int> Numbers { get; }
        }

        [Fact]
        public void EqualRecordsShouldBeEqual()
        {

            var a = new Foo( 10, 20, List( 0, 1, 2 ) );
            var b = new Foo( 10, 20, List( 0, 1, 2 ) );

            a.Equals( b ).Should().Be( true );

        }

        [Fact]
        public void NonEqualRecordsShouldBeNotEqual()
        {

            var a = new Foo( 10, 20, List( 0, 1, 2 ) );
            var b = new Foo( 10, 20, List( 0, -1, 2 ) );

            a.Equals( b ).Should().Be( false );

        }
    }
}

```

and it fails with the following error

> System.TypeInitializationException : The type initializer for 'LanguageExt.RecordType`1' threw an exception.
> ---- System.ArgumentException : 'Default' is not a member of type 'Utils.Spec.LanguageExtRecords+Foo'
>    at LanguageExt.Record`1.Equals(RECORDTYPE other)
>    at Utils.Spec.LanguageExtRecords.EqualRecordsShouldBeEqual()
> ----- Inner Stack Trace -----
>    at System.Linq.Expressions.Expression.PropertyOrField(Expression expression, String propertyOrFieldName)
>    at LanguageExt.IL.<>c__DisplayClass10_0`1.<<GetHashCode>g__Fields2>d.MoveNext()
>    at LanguageExt.List.fold[S,T](IEnumerable`1 list, S state, Func`3 folder)
>    at LanguageExt.IL.GetHashCode[A]()
>    at LanguageExt.RecordType`1..cctor()

It looks like it is mistakenly thinking the **Default** is an instance property.

bradphelan:
Fixes #281

### DIFF
--- a/LanguageExt.Core/Utility/Reflect.cs
+++ b/LanguageExt.Core/Utility/Reflect.cs
@@ -30,7 +30,7 @@ namespace LanguageExt
 #if !COREFX13
                                     .OrderBy(p => p.MetadataToken)
 #endif
-                                    .Where(p => p.CanRead && p.GetMethod.IsPublic && !p.GetAccessors(true)[0].IsStatic)
+                                    .Where(p => p.CanRead && p.GetMethod.IsPublic && !IsStatic(p))
                                     .Where(p => !toSet(p.CustomAttributes.Map(a => a.AttributeType.Name)).Intersect(excludeAttrsSet).Any())
                                     .ToArray();
 
@@ -46,6 +46,18 @@ namespace LanguageExt
 
             return Enumerable.Concat(publicFields, backingFields);
         }
+
+        /// <summary>
+        /// Returns true if the property is static by inspecting
+        /// the static property of the accessors. Note that if
+        /// there are no accessors then the property is assumed
+        /// to be **non static**. Not sure that this case is
+        /// even possible in CLR.
+        /// </summary>
+        /// <param name="p"></param>
+        /// <returns></returns>
+        public static bool IsStatic
+            (PropertyInfo p) => p.GetAccessors( true ).HeadOrNone().Map( x => x.IsStatic ).IfNone( false );
 
         public static Option<MethodInfo> GetPublicStaticMethod(Type type, string name, Type argA) =>
             type.GetTypeInfo()
--- a/LanguageExt.Core/Utility/Reflect.cs
+++ b/LanguageExt.Core/Utility/Reflect.cs
@@ -30,7 +30,7 @@ namespace LanguageExt
 #if !COREFX13
                                     .OrderBy(p => p.MetadataToken)
 #endif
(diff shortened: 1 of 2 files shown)


bradphelan:
If this could be merged  quickly I would appreciate it as it's blocking my integration work and there is no easy work around.

LanguageExt.Core/Utility/Reflect.cs
```diff
@@ -30,7 +30,7 @@ public static IEnumerable<FieldInfo> GetPublicInstanceFields<A>(params Type[] ex
 #if !COREFX13
                                     .OrderBy(p => p.MetadataToken)
 #endif
-                                    .Where(p => p.CanRead && p.GetMethod.IsPublic)
+                                    .Where(p => p.CanRead && p.GetMethod.IsPublic && !p.GetAccessors(true)[0].IsStatic)
```
louthy: Could `GetAccessors` ever return an empty array?  Would it be better to call: 

`!p.GetAccessors(true).HeadOrNone().Map(x => x.IsStatic).IfNone(true)`

Because then if `GetAccessors` ever did return an empty array then the property would be filtered out.

bradphelan: It can return zero

image

https://msdn.microsoft.com/en-us/library/ff3b842x(v=vs.110).aspx

will fix.

bradphelan: Actually.. When you pass **true** to GetAccessors then the array will always have two elements

https://msdn.microsoft.com/en-us/library/afadtey7(v=vs.110).aspx

> An array whose elements reflect the get and set accessors of the property reflected by the current instance. **If nonPublic is true, this array contains public and non-public get and set accessors.** If nonPublic is false, this array contains only public get and set accessors. If no accessors with the specified visibility are found, this method returns an array with zero (0) elements.

Still just to be safe I'll fix the code

louthy:
@bradphelan Thanks Brad, looks good 👍

PR #287, opened 2017-10-19
Fix equality for Arr<A>

### LINKED ISSUE #286: Stackoverflow in Arr<Arr<double>>.Equals
image

The test case that generates the stack overflow is

        [Fact]
        public void ArrShouldNotStackOverflowOnEquals()
        {
            Arr<Arr<double>> arr = default;

            arr.Equals( arr ).Should().BeTrue();
        }

Wierd!!

bradphelan:
Fixes #286 

There probably should be more testing on equality for all the types. This just fixes ``Arr<T>``

### DIFF
--- a/LanguageExt.Core/DataTypes/Arr/Arr.cs
+++ b/LanguageExt.Core/DataTypes/Arr/Arr.cs
@@ -449,6 +449,22 @@ namespace LanguageExt
         public override bool Equals(object obj) =>
             obj is Arr<A> && Equals((Arr<A>)obj);
 
+        [Pure]
+        public bool Equals(Arr<A> obj) =>
+            SafeEquals(Value, obj.Value);
+
+
+        private static bool SafeEquals(A[] strA, A[] strB)
+        {
+            var length = strA.Length;
+            if (length != strB.Length)
+                return false;
+            for (var i = 0; i < length; i++)
+                if (!strA[i].Equals( strB[i] ))
+                    return false;
+            return true;
+        }
+
         /// <summary>
         /// Get the hash code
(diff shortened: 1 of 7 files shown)


LanguageExt.Core/DataTypes/Arr/Arr.cs
```diff
@@ -449,6 +449,22 @@ public Arr<B> Map<B>(Func<A, B> map)
         public override bool Equals(object obj) =>
             obj is Arr<A> && Equals((Arr<A>)obj);
 
+        [Pure]
+        public bool Equals(Arr<A> obj) =>
+            SafeEquals(Value, obj.Value);
```
louthy: Should probably just call `default(MArr<A>).Equals(this, obj);`

bradphelan:
doesnt that allocate and then deallocate a backing array every time you do
that?

louthy:
@bradphelan 

> doesnt that allocate and then deallocate a backing array every time you do
that?

Nope. `MArr<A>` is the 'class instance' for the `Arr<A>` data-type, it doesn't allocate anything, just calls: `Enumerable.SequenceEqual(x, y)`

louthy:
@bradphelan Added my own fixes for this, so closing.  Thanks for the heads up (Y).

bradphelan:
no i mean that Arr<T> is a struct but is backed by an array reference which
needs to allocate on heap. Therefore default(Arr<T>) will allocate an array
every time it is called.

louthy:
It's default(MArr) not default(Arr) - the Monad/Eq/Ord type-class instance for Arr.

bradphelan:
my bad! thanks for the clarification

PR #288, opened 2017-10-19
Use gitversion for versioning

bradphelan:
Instead of patching the csproj file I use [GitVersion](http://gitversion.readthedocs.io/en/latest/) and pass the version number directly to the ``dotnet pack`` command. This makes creating pre-release packages easier for contributers. For example now when I build for the current branch I get SemVer 2.0 compatible names and on every commit the version is incremented.

```
C:\Users\phelan\workspace\language-ext [easy-versioning +2 ~5 -0 | +0 ~2 -0 !]> ls ..\artifacts\bin\

    Directory: C:\Users\phelan\workspace\artifacts\bin

Mode                LastWriteTime         Length Name
----                -------------         ------ ----
-a----       10/19/2017  10:58 AM        3193099 LanguageExt.Core.2.1.24-easy-versioning.17.nupkg
-a----       10/19/2017  10:58 AM          10214 LanguageExt.FSharp.2.1.24-easy-versioning.17.nupkg
-a----       10/19/2017  10:58 AM         133303 LanguageExt.Parsec.2.1.24-easy-versioning.17.nupkg
```
Updating the next real version is as simple as editing the GitVersion.yml file
```
C:\Users\phelan\workspace\language-ext [easy-versioning +2 ~5 -0 ~]> cat .\GitVersion.yml
mode: ContinuousDeployment
next-version: 2.1.24
branches: {}
ignore:
  sha: []
```

### DIFF
--- a/GitVersion.yml
+++ b/GitVersion.yml
@@ -1,0 +1,5 @@
+mode: ContinuousDeployment
+next-version: 2.1.24
+branches: {}
+ignore:
+  sha: []
--- a/LanguageExt.Core/LanguageExt.Core.csproj
+++ b/LanguageExt.Core/LanguageExt.Core.csproj
@@ -12,7 +12,6 @@
     <DefineConstants>COREFX13</DefineConstants>
   </PropertyGroup>-->
   <PropertyGroup>
-    <PackageVersion>2.1.22</PackageVersion>
     <PackageId>LanguageExt.Core</PackageId>
     <Title>LanguageExt.Core</Title>
     <Authors>Paul Louth</Authors>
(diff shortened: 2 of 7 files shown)


louthy:
I'll need a few days to take this in, I'm not massively keen on changing my deployment process at the moment, because it's pretty seamless which means I tend to get things deployed quickly.

bradphelan:
This would hopefully make it more **seamless** as you don't have to rely on AssemblyInfo.cs patching to set the version number, rather you can just just use msbuild and pass the version property in. That I'm using **gitversion** to generate the version number is just an extra feature. You could just as easily have the version number in a text file and set as an environment variable and pass that to msbuild. However gitversion makes it so that you can't forget to bump the (CI) version number on every checkin.

louthy:
Going to close this, as I'm not really looking to change my deployment process right now

PR #289, opened 2017-10-19
Bug in biIter for Option / OptionUnsafe (fix attached)

StefanBertels:
The attached fix results in correct behaviour for  Option(Unsafe).BiIter(...)

call fa when some
call fb when none

A small test is included.


louthy:
Thanks @StefanBertels 👍

Issue #283, opened 2017-10-18
Converting async computations to Try


cfiet:
Hi,

I'm having simillar problem to https://github.com/louthy/language-ext/issues/255, trying to convert async code to exception-safe Try.

I see the issue with the @trbngr suggestion though, since his code roughly translates to:

```
        public async Task<IActionResult> GetGendersAsync()
        {
            var result = _repository.GetAllGendersAsync()
            return Try(await result).Match(...);
        }
```

The third line can still throw any exception which will not be captured by Try and will just bubble up the stack.

I can't really find any way of constructing Try or TryAsync with Func<Task<T>> so it only continues only when the Task is completed/failed/cancelled. Is there any helper method that allows to convert Func<Task<T>> to Task<Try<T>>? Something along the line of:
```
    public async Task<Try<TValue>> TaskToTry<TValue>(Func<Task<TValue>> asyncComputation)
    {
       try
       {
          return Try(await asyncComputation());
       }
       catch(Exception e)
       {
          return Try<TValue>(e);
       }
    }
```

trbngr:
You summoned me.

I thought I'd give this a shot to my suprise, there was no obvious answer. It took some digging into the source.

I then saw [this](https://github.com/louthy/language-ext/blob/master/LanguageExt.Core/DataTypes/Try/Task.Try.Extensions.cs#L32)

Which led me to write a [test](https://github.com/trbngr/langext-issue255/blob/master/langext-issue255/UnitTest1.cs#L76)

TLDR: use the `.ToAsync()` extension off of `Try<Task<A>>` to turn it into a `TryAsync<A`.

Enjoy ;)

cfiet:
Hi,

Thank you for your quick answer, your solution seems to be working in almost every case. I did find one test case where it fails, see the second test case [in this file](https://github.com/cfiet/langext-issue255/blob/issue283/langext-issue255/issue283.cs). I think this is due to a bug/missing specialised definition for ToAsync for untyped tasks. I'm working on addint the missing declaration.

The whole thing left me with a question of what is the purpose of TryAsync type? All it does is it runs computation via Task.Run, but does not really seem to lift/unlift the task correctly, so it can't be used with async methods and Tasks. Any insight here?

louthy:
@cfiet Your second example doesn't make sense:
```c#
        public async Task SyncMethodReturingTaskShouldResultInFail()
        {
            var result = await Try(() => Task.FromException(new Exception())).ToAsync().Match(
[...]

PR #290, opened 2017-10-19
Added missing ToAsync implementation

cfiet:
Hi, 

First of all, thanks for this amazing library.

This pull request solvs some problems I've encountered related to issue #283.

Changes:
1. Added missing implementation of ToAsync that handles untyped Tasks,
2. Removed unnecessary check of Task status - awaiting on task.Value will result in exception being thrown if underlying task is faulted or cancelled, so there's no need to manually check if task has faulted,
3. Added tests for ToAsync to verify the changes


louthy:
Looks good, thanks!

PR #294, opened 2017-10-24
[Don't Merge] Get ordering working for M<N<T>>

### LINKED ISSUE #293: How to get Ord<Record<T>> working when members are Set<U> or Arr<U>
I have the following test

```
    public class ContainerOrderingTests
    {
        public class WithSet : Record<WithSet>
        {
            public WithSet(int i, Set<int> s)
            {
                I = i;
                S = s;
            }

            public int I { get; }
            public Set<int> S { get; }
        }

        [Fact]
        public void SetOrderingWorks()
        {
            var a = new WithSet(0, Prelude.Set(0,1,2));
            var b = new WithSet(0, Prelude.Set(0,1,2));
            Assert.Equal(0,a.CompareTo( b )) ;
        }
    }

```

which fails with the fact that Set does not implement IComparable. But somehow I'm sure you have thought of this but the solution is non obvious.

bradphelan:
This pull request is just for tracking progress and to allow commits from @louthy if he likes

Fixes #293 

With these changes **Ord** classes can be found for containers that match patterns

   M<N<T>>

which 

   Ord<Set<int>>
   Ord<Lst<string>>

for example do.

TODO

- [ ] Improve performance
- [ ] Generalize
- [ ] Add more test cases to catch problems

### DIFF
--- a/LanguageExt.Core/ClassInstances/Class.cs
+++ b/LanguageExt.Core/ClassInstances/Class.cs
@@ -46,6 +46,14 @@ namespace LanguageExt.ClassInstances
                 return;
             }
 
+            All = toSet<OrdTypeInfo,TypeInfo>( GetOrderableContainerOrd2().Select(v=>v.GetTypeInfo()));
+
+            if(All.Count == 1)
+            {
+                Default = (A)Activator.CreateInstance(All.Head().AsType());
+                return;
+            }
+
             //if(All.Count == 0)
             //{
             //    Default = TryHigherKind();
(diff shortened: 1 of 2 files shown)


bradphelan:
I'm thinking Ninject is the way to solve these binding problems

https://stackoverflow.com/questions/6740992/ninject-how-to-bind-an-open-generic-with-more-than-one-type-argument

bradphelan:
No. Ninject can't do it. Maybe 

http://simpleinjector.readthedocs.io/en/latest/advanced.html#generics

bradphelan:
No SimpleInjector can only inject class types not value types :(

bradphelan:
I've noticed a problem with the libraries. We have for example ``Set<A>`` and ``Set<OA,A>`` but we only have ``OrdSet<OA,A>`` which is defined as 

    /// <summary>
    /// Equality and ordering
    /// </summary>
    public struct OrdSet<ORD, A> : Ord<Set<A>>
        where ORD : struct, Ord<A>
    {


but not 

    /// <summary>
    /// Equality and ordering
    /// </summary>
    public struct OrdSet<ORD, A> : Ord<Set<ORD,A>>
        where ORD : struct, Ord<A>
    {

which seems like there is a missing part of the puzzle somewhere.

Issue #244, opened 2017-07-28
How to upgrade Rws<>?


gwintering:
What is the status ReaderWriterState in the Type Classes rewrite?  Was support eliminated in 8cff584, or is it now just a special case of stacked monads that Transformers can handle?  I ask because I have a process that uses the old `Rws<>` and I'm unsure how to rewrite it for v2 of LanguageExt.  I'd gladly contribute documentation or tests if you can point me in the right direction.

Thank you so much for this library.  It's wonderful how it allows me to gradually inculcate functional features into my organization without the culture shock of switching languages.  And I've learned a ton about functional programming just by reading the code and documentation.

louthy:
@gwintering Hi Greg, I dropped support for it because I genuinely thought nobody would use it due to it being generics hell!  

I'd be up for re-adding it, but I would want to generalise it properly like I have done with the `Writer` (to use monoids), with a more specific version that mimics the original.  I think the last time I tried it it turned my head inside out, so rather than commit something buggy I decided to drop it.

This is the old source if you want to build a version for yourself:

```c#
using LanguageExt;
using System;
using System.Collections.Generic;
using System.ComponentModel;
using System.Diagnostics.Contracts;
using System.Linq;
using System.Text;
using System.Threading.Tasks;
using static LanguageExt.Prelude;

namespace LanguageExt
{
    /// <summary>
    /// Combines all the behaviours of the Reader, Writer and State monads into a single
    /// monadic computation.  
    /// </summary>
    /// <typeparam name="R">Reader type</typeparam>
    /// <typeparam name="W">Writer type</typeparam>
    /// <typeparam name="S">State type</typeparam>
    /// <typeparam name="T">Wrapped type</typeparam>
    /// <param name="rws">A tuple (environment, state) that represents the shared immutable
    /// environment and shared mutable state.</param>
    /// <returns>A value that represents the bound value as well as the shared environment,
    /// shared state, and shared outputs</returns>
    public delegate RwsResult<W, S, T> Rws<R, W, S, T>(Tuple<R, S> rws);

    /// <summary>
    /// RWS result.
    /// </summary>
    public struct RwsResult<W, S, T>
    {
        readonly T value;
        public readonly IEnumerable<W> Output;
        public readonly S State;
        public readonly bool IsBottom;

        internal RwsResult(IEnumerable<W> output, S state, T value, bool isBottom = false)
        {
            this.value = value;
            Output = output;
            State = state;
            IsBottom = isBottom;
        }

[...]

PR #300, opened 2017-11-06
Reintroduce RWS Monad

gwintering:
Hi Paul,

I finally got enough time to finish this up and review it.

I've now got a much deeper respect for your efforts in writing this library; there are quite a lot of API functions for each monad.

I'm still interested in your `MType` idea or any other way to hide the type parameter noise but that's probably a separate issue.

#244

### DIFF
--- a/LanguageExt.Core/ClassInstances/Monad/MRWS.cs
+++ b/LanguageExt.Core/ClassInstances/Monad/MRWS.cs
@@ -1,0 +1,126 @@
+using System;
+using LanguageExt.TypeClasses;
+using static LanguageExt.TypeClass;
+using System.Diagnostics.Contracts;
+using static LanguageExt.Prelude;
+using System.Threading.Tasks;
+
+namespace LanguageExt.ClassInstances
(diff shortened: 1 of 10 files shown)


louthy:
Hi Greg, thanks for this, I'll do my best to review this week

louthy:
@gwintering Hi Greg, fabulous stuff!  I'd probably prefer a few more tests and comments, but, to be honest with you, I don't mind too much as I understand how much just writing this stuff can turn your head inside out!  Thanks for the effort 👍 

> I've now got a much deeper respect for your efforts in writing this library; there are quite a lot of API functions for each monad.

Yep, I wish it could be simpler, but this the best approach I can think of that is type-safe.

> I'm still interested in your MType idea or any other way to hide the type parameter noise but that's probably a separate issue.

I'll need to revisit that.  But I think, when we spoke about it originally, I tried some prototype ideas that didn't really work out.  I'm very time poor at the moment, so I'll have to come back to that later.

Paul

PR #302, opened 2017-11-06
Additional map constructor functions

OlduwanSteve:
With two strategies for handling duplicate keys.


louthy:
Thanks @OlduwanSteve

Issue #304, opened 2017-11-06
Ambiguous type in List/Seq.foldX() when ListItem type T and State Type S are same types.


mrgchr:
Hello,

I'm not sure that current behavior is by design or not.

List(or Seq).foldX(foldWhile, foldBackWhile, foldUntil, foldBackUntil) makes compile time error when type S and T are same types.
A compilation time error will happen when you build following code.

```
var input = Seq(10, 20, 30, 40, 50);
var output = foldBackWhile(input, 0, (s, x) => s + x, x => x >= 40);
//The call is ambiguous between the following methods or properties:
//'Seq.foldBackWhile<S, T>(Seq<T>, S, Func<S, T, S>, Func<T, bool>)' and
//'Seq.foldBackWhile<S, T>(Seq<T>, S, Func<S, T, S>, Func<S, bool>)'
```

# Suggestion

If this behavior is not by design, I would like to suggest following modification(and I'm ready to make a PR).

Renaming `pred` argument(3rd one) to `preditem` and `predstate` in order to add explicitly named argument hint to compiler.

```
+ public static S foldWhile<S, T>(Seq<T> list, S state, Func<S, T, S> folder, Func<T, bool> preditem);
+ public static S foldWhile<S, T>(Seq<T> list, S state, Func<S, T, S> folder, Func<S, bool> predstate);

- public static S foldWhile<S, T>(Seq<T> list, S state, Func<S, T, S> folder, Func<T, bool> pred);
- public static S foldWhile<S, T>(Seq<T> list, S state, Func<S, T, S> folder, Func<S, bool> pred);
```

usage:
```
var input = Seq(10, 20, 30, 40, 50);

// In this case, preditem(x) is ListItem (int)
var output3 = foldBackWhile(input, 0, (s, x) => s + x, preditem: x => x >= 40);
Assert.Equal(90, output3);

// In this case, predstate is State (int)
var output4 = foldBackWhile(input, 0, (s, x) => s + x, predstate: s => s < 90);
Assert.Equal(90, output4);
```

I know that modifying argument names could be breaking code modification.
However, if you are OK with this, I would like to make a PR.

Regards.

louthy:
Rolled out to v2.1.26

PR #305, opened 2017-11-08
Avoid ambiguous calling when S and T are same type

mrgchr:
Fix of #304 .

Hello,

This PR is in order to avoid ambiguous method calling and add related unit tests.
List(or Seq).foldX(foldWhile, foldBackWhile, foldUntil, foldBackUntil) causes compile time error when type S and T are same types.
A compilation time error will happen when you build following code.

```
var input = Seq(10, 20, 30, 40, 50);
var output = foldBackWhile(input, 0, (s, x) => s + x, x => x >= 40);
//The call is ambiguous between the following methods or properties:
//'Seq.foldBackWhile<S, T>(Seq<T>, S, Func<S, T, S>, Func<T, bool>)' and
//'Seq.foldBackWhile<S, T>(Seq<T>, S, Func<S, T, S>, Func<S, bool>)'
```


louthy:
Thanks 👍

PR #307, opened 2017-11-20
Option.AsEnumerable returns Seq<T>

mrgchr:
Hello,

I have noticed that Option.AsEnumeragle() returns Seq<T> type. (so OptionUnsafe, OptionAysnc are)
I'm not sure this is by design or not.

This PR includes that the fix of type declaration and unit tests related to fix.


louthy:
It's by design.  Seq is a better IEnumerable for short sequences

mrgchr:
Thank you.

PR #309, opened 2017-11-21
Error handling exceptions in OptionAsync Match

ggmueller:
Hi,

it seems like there is a problem with Exceptions in a Match, when its thrown inside an async Action.
The unit test attached reproduces the problem.

Running in an application this problem also kills the problem, as it causes an unhandled Exception on the ThreadPool.

Regards Georg

### DIFF
--- a/LanguageExt.Tests/OptionAsyncTests.cs
+++ b/LanguageExt.Tests/OptionAsyncTests.cs
@@ -115,5 +115,29 @@ namespace LanguageExt.Tests
 
             taskOpt = optTask.Sequence();
         }
+
+        [Fact]
+        public async void HandlesException ()
+        {
+            Option<int> i = 1;
+            
+            await Assert.ThrowsAsync<MyException>(async () => await i.ToAsync().Match (
+                    async n =>
+                    {
+                        await DoWork();
+                        throw new MyException (n.ToString());
+                    },
+                    () => { }));
+        }
+        
+        
+
+    }
+
+    public class MyException : Exception 
+    {
+        public MyException (string message):base(message)
+        {
+        }
     }
 }


louthy:
You talk about 'the problem' without explaining what your problem is.
Please be explicit

ggmueller:
In this case, the exception is not resolved to a Faulted Task, but instead kills the ThreadPool.
Thus code outside of the Match cannot handle the exception.

mrgchr:
Excuse me for cutting in. I'm also interested in this case.

In my env, following test case fails:
```cs
[Fact]
public async void HandlesException()
{
    var i = Some(1);

    await Assert.ThrowsAsync<DivideByZeroException>(async () =>
        await i.ToAsync().Match(async n =>
            {
                await Task.Delay(10);
                var zero = 0;
                var result = n / zero;
            },
            () => { }));
}
```

But, following code passes:
```cs
[Fact]
public async void HandlesException()
{
    var i = Some(1);

    await Assert.ThrowsAsync<DivideByZeroException>(async () =>
        await i.ToAsync().Match(async n =>
            {
                await Task.Delay(10);
                var zero = 0;
                var result = n / zero;
                return result;
            },
            () =>
            {
                return 0;
            }));
}
```

So, following code is almost same as the original test case and pass successfully:
```cs
[Fact]
public async void HandlesException()
{
    var i = Some(1);

    await Assert.ThrowsAsync<DivideByZeroException>(async () =>
        await i.ToAsync().Match(async n =>
            {
                await Task.Delay(10);
                var zero = 0;
                var result = n / zero;
                return Unit.Default;
            },
            () =>
            {
                return Unit.Default;
            }));
}
```

I have noticed that original test case is uging `Match(Action<A> Some, Action None` in OptionAsync.cs.
In my personal opinion, I guess following new overload methods might solve this type of case for example:

```cs
        public Task<Unit> Match(Func<A, Task> Some, Func<Task> None) =>
            MOptionAsync<A>.Inst.MatchAsync(this, Some, None);

        public Task<Unit> Match(Func<A, Task> Some, Action None) =>
            MOptionAsync<A>.Inst.MatchAsync(this, Some, None);

        public Task<Unit> Match(Action<A> Some, Func<Task> None) => ...
```
```cs
        public async Task<Unit> MatchAsync(OptionAsync<A> opt, Func<A, Task> Some, Func<Task> None)
        {
            if (Some == null) throw new ArgumentNullException(nameof(Some));
            if (None == null) throw new ArgumentNullException(nameof(None));
            if (await opt.IsSome) await Some(await opt.Value); else await None();
            return Unit.Default;
        }

        public async Task<Unit> MatchAsync(OptionAsync<A> opt, Func<A, Task> Some, Action None)
        {
            if (Some == null) throw new ArgumentNullException(nameof(Some));
            if (None == null) throw new ArgumentNullException(nameof(None));
            if (await opt.IsSome) await Some(await opt.Value); else None();
            return Unit.Default;
        }
```

Again, sorry for cutting in. I hope this post is helpful.

louthy:
This has been dealt with on the async overhaul.

ggmueller:
I will test that in the new year. New release looks good though. 

Thanks for your great support. Its a real pleasure working with language-ext.

PR #319, opened 2017-12-22
moved type parameter documentation from a method where they are used …

bender2k14:


louthy:
@bender2k14 What's the justification for this?  Visual Studio injects the type-params there by default (and not on the method).  It would be an enormous job to change this for the whole project.  So for consistency it seems better to leave them where they are, because the type-params are type arguments, not method arguments.  

I think it's especially true for `Functor` which is a 'class-instance' - i.e. the type is pretty much part of the method definition too.

bender2k14:
I was just trying to change something that I thought was wrong.  I don't know what Visual Studio injection you are talking about.

If (hypothetically) in Visual Studio, you refactor-rename these generic type parameters with the documentation as is, then the documentation will not be updated to match.  If instead the documentation moved to where I put it, then the documentation will also be changed to match.

I thought I was making things more consistent.  Of the four generic interfaces in the TypeClasses namespace that I have looked at, the majority documented each generic type parameter defined on type on the defining type (as I my change does with Functor.cs).

- All generic type parameters defined on types in these files are documented on its defining type
- - Applicative.cs
- - Monad.cs
- - Monoid.cs
- All generic type parameters defined on types in these files are documented on a method
- - Functor.cs (but my change makes this files consistent with those in the previous group)
- All generic type parameters defined on types in these files are missing documentation
- - Semigroup.cs

As both my first pull request to your project and also first pull request on this type of change, I wanted to just modify one small file to get feedback from you first before spending time making similar changes in many files.

It may very well be an enormous job to make this type a change for the entire project.  That doesn't bother me though.  I enjoy making types of improvements.  I have been reading the documentation to gain a better understanding of the vocabulary from functional programming.  Doing this with your library is rather ideal because C# is my best language.  Furthermore, I am interested in aspects of quality assurance, and might use some tool to identify possible changes like this one and ideally even ones that can make the change for me.

louthy:
So sorry, I just realised I was arguing the wrong way around.  Yes, it's wrong that they're attached to the method.  For some reason my quick scan of the changes gave me the impression you were moving the typeparams _to_ the method, not _from_ it to the type.

This is almost certainly a legacy of when I was trying various approaches to the type-class system and the generic params were part of the method.

PR #322, opened 2017-12-24
improve null checking

bender2k14:
I refactored the tests for `object` extension method `IsNull` and `Prelude` methods `isnull` and `notnull` so that each test includes exactly one assert while also not duplicating any code.  I added more tests, especially tests for `null` and not `null` values of a `Nullable<>` type.  Then I slightly improved the readability of the implementation of `Check<T>.IsNull`.

### DIFF
--- a/LanguageExt.Tests/NullChecks/AbstractNullCheckTests.cs
+++ b/LanguageExt.Tests/NullChecks/AbstractNullCheckTests.cs
@@ -1,0 +1,43 @@
+﻿using Xunit;
+
+namespace LanguageExt.Tests.NullChecks
+{
+    public abstract class AbstractNullCheckTests
+    {
+        protected abstract bool ExpectedWhenNull { get; }
+        protected abstract bool NullCheck<T>(T value);
+
+        protected bool ExpectedWhenNotNull => !ExpectedWhenNull;
+
+        [Fact]
+        public void NullCheck_NullString_AsExpectedWhenNull()
+        {
+            string value = null;
+
+            var actual = NullCheck(value);
+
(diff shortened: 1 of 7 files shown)


LanguageExt.Tests/NullChecks/AbstractNullCheckTests.cs
```diff
@@ -0,0 +1,127 @@
...
+        protected abstract bool ExpectedWhenNull { get; }
+        protected abstract bool NullCheck<T>(T value);
+
+        private bool ExpectedWhenNotNull => !ExpectedWhenNull;
+
+        [Fact]
+        public void NullCheck_NullObject_AsExpectedWhenNull()
+        {
+            object value = null;
+
```
louthy: This commit is all great, but could you do me a favour and remove the empty-line every other line:

So this:
```c#
object value = null;

var actual = NullCheck(value);

Assert.Equal(ExpectedWhenNull, actual);
```
Becomes:
```c#
object value = null;
var actual = NullCheck(value);
Assert.Equal(ExpectedWhenNull, actual);
```
Just as a general rule I like to keep related blobs of code together, only spacing when there's a natural 'pause' in what's happening - the message to the programmer being "We've done that, now we're doing this".  I think because the setup/test/assert are one-liners here, there's no real "we're doing the setup bit, now we're moving to the testing bit, ..."

bender2k14: Sure thing.

louthy:
Thanks, that's great stuff 👍

PR #323, opened 2017-12-25
Feature/improve default value checking

bender2k14:
I refactored the tests for `object` extension method `IsDefault` and `Prelude` methods `isDefault` and `notDefault` so that each test includes exactly one assert while also not duplicating any code. I added more tests, especially tests for `null` and not `null` values of a `Nullable<>` type. Then I redefined `Prelude.isDefault` via `object` extension method `IsDefault`.


louthy:
Good stuff 👍

PR #324, opened 2017-12-25
moved None from Prelude to Option.Prelude

bender2k14:
Moved the location of `public static OptionNone None` from `Prelude.cs` to `Option.Prelude.cs`.

### DIFF
--- a/LanguageExt.Core/DataTypes/Option/Option.Prelude.cs
+++ b/LanguageExt.Core/DataTypes/Option/Option.Prelude.cs
@@ -66,6 +66,12 @@ namespace LanguageExt
             value.IsNone;
 
         /// <summary>
+        /// 'No value' state of Option T.
+        /// </summary>
+        public static OptionNone None =>
+            OptionNone.Default;
+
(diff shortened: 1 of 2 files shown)


louthy:
Rejecting this.  The point of the Prelude is to have a single static type that devs can include to get the constructors for all the key types.  This makes discovery harder and will break everybody’s builds.

bender2k14:
I must be missing something.  How would this break anyone's build?  I moved a method from one partial implantation of the `Prelude` type to a different partial implementation of the `Prelude` type.  I don't see any problem with that.  The `Option<>` type has three different names for constructors: `None`, `Some`, and `Optional`.  What this PR does is move the `None` constructor to the same file where the `Some` and `Optional` constructors are located.

louthy:
Apologies, reading on my phone - probably didn’t look properly.  I’ll take a look tomorrow when the ‘Christmas cheer’ has worn off ;)

louthy:
I've read it properly now.  Thanks 👍

PR #325, opened 2017-12-25
Feature/reorganize null checking code

bender2k14:
Rearranged the decencies among the null checking methods.  Now, `Prelude.notnull` depends on `Prelude.isnull` which depends on the `IsNull` extension method.  My motivation for this change is to move some complexity from the higher complexity file `Prelude.cs` to to the lower complexity file `ObjectExt.cs`.


bender2k14:
Ok, that makes sense.  Reverted the type name change.

bender2k14:
May I ask why this was closed?

louthy:
Sorry, I thought it was just the name change.  I see you've moved the Check type now.

PR #330, opened 2017-12-29
utilzing the Check class to implement IsDefault extension method

bender2k14:
The extension method `IsDefault` is now defined via the `Check<T>` class.

### DIFF
--- a/LanguageExt.Core/Extensions/ObjectExt.cs
+++ b/LanguageExt.Core/Extensions/ObjectExt.cs
@@ -45,14 +45,12 @@ namespace LanguageExt
 
         private static class Check<T>
         {
-            static readonly T DefaultT;
             static readonly bool IsReferenceType;
             static readonly bool IsNullable;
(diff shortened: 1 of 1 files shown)


LanguageExt.Core/Extensions/ObjectExt.cs
```diff
@@ -45,22 +45,28 @@ public static class ObjectExt
...
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
+            internal static bool IsDefault(T value) =>
+                DefaultEqualityComparer.Equals(value, DefaultT);
+
             [MethodImpl(MethodImplOptions.AggressiveInlining)]
             internal static bool IsNull(T value) =>
                 IsNullable
-                    ? value.Equals(default(T))
-                    : IsReferenceType && DefaultEqualityComparer.Equals(value, default(T));
+                    ? value.Equals(DefaultT)
```
louthy: This should stay as `default(T)`, it's more efficient than accessing the `DefaultT` field.

PR #332, opened 2017-12-30
make random actually thread safe

bender2k14:
Both `Prelude.random` overloads claimed to be thread safe, but one of them was not, making neither thread safe.  Fixed this problem.


louthy:
Thanks 👍

Issue #336, opened 2017-12-31
Missing extensions call TryGetValue returning Option<>


bender2k14:
In main documentation for this project at the end of the [section about `out` parameters](https://github.com/louthy/language-ext#the-awful-out-parameter), it says that extension methods exists for
- `IImmutableDictionary<K, V>.TryGetValue`
- `IImmutableSet<K, V>.TryGetValue`

that return `Option<V>`.  However, I can't find these methods anywhere.  The other two extension named `TryGetValue` that return `Option<>` are defined in [`TryOutExt.cs`](https://github.com/louthy/language-ext/blob/master/LanguageExt.Core/Extensions/TryOutExt.cs).

Would you like these extensions added to match the documentation or change the main documentation to match the code?

louthy:
They’re dead, I replaced dropped support for the ImmutableCollections, so
that can go

PR #338, opened 2018-01-01
removed documentation for two nonexistent TryGetValue extensions

bender2k14:
As title says, which fixes issues #336.


louthy:
👍

Issue #341, opened 2018-01-08
Result<>.Bottom.IsBottom is false


oliverhanappi:
I would expect the expression `Result<object>.Bottom.IsBottom` to return `true`, however it returns `false`. I do not know if this is by design or not, however I think it is pretty unobvious.

louthy:
It’s a bug, I’ll take a look when I can.

PR #343, opened 2018-01-10
Fix bug341 explicit bottom result is bottom

bender2k14:
This PR claims to fix issue #341.

The new behavior makes `Result<object>.None` and `Result<object>.Bottom` the same.


louthy:
Thanks for the commit, I took a slightly different path.

bender2k14:
Welcome, and no worries about not using this code.  I am interested to see how you fixed it.  Have you pushed your fix somewhere?

bender2k14:
I see.  It is included in commit c9bae2d41d0de614100d7cf1480625774a8bfd0a.

Issue #346, opened 2018-01-19
NullReferenceException due to Seq<T>, SeqEmpty<T> static initializers in .Net Framework


chelxom:
This issue happens in .Net Framework (my testing environment is netfx461) but not in .Net Core.

### Repro
```C#
using LanguageExt;
using System.Linq;

namespace SeqStaticInitializer.NetFramework
{
    class Program
    {
        static void Main(string[] args)
        {
            var list = 1.Cons().ToList();
        }
    }
}
```
It's with source in master branch. Such code throws `NullReferenceException` that `Seq<int>.Empty` is `null` in .Net Framework.

I think the reason might be that if there's no static constructor, static field initialization is implementation-dependent([C# spec v5, 10.5.5.1](https://www.microsoft.com/en-us/download/details.aspx?id=7029)). And initialization of `Seq<T>.Empty` and `SeqEmpty<T>.Default` is in unclear order due to their type hierarchy.

louthy:
The code works for me on .NET 4.6.1.  Please make sure you're on the latest build - as I can't see how static field initialization would have any bearing on this issue.  

Please comment or re-open if you still have issues.

PR #347, opened 2018-01-22
Added Null check for calling ToList on Single Element Cons

jeremysmith1:
This fixes issue: 
https://github.com/louthy/language-ext/issues/346 

I added a test as well.


jeremysmith1:
Not sure how I did that to be honest. Fixing.

PR #353, opened 2018-02-07
Memoize Tasks for Async Tries

### LINKED ISSUE #345: Improve Memo for TryAsync, TryOptionAsync
The `Memo` implementations for both `TryAsync` and `TryOptionAsync` cache the  task _result_, which is wrapped up into a monad on subsequent calls.  But tasks already cache their results, so we should cache the  _task_ as soon as it's available and return it on subsequent calls.  This way we can provide the memoized value sooner and avoid unnecessary `Try` calls.

gwintering:
closes #345

### DIFF
--- a/LanguageExt.Core/DataTypes/TryAsync/TryAsync.Extensions.cs
+++ b/LanguageExt.Core/DataTypes/TryAsync/TryAsync.Extensions.cs
@@ -22,11 +22,11 @@ public static class TryAsyncExtensions
     public static TryAsync<A> Memo<A>(this TryAsync<A> ma)
     {
         bool run = false;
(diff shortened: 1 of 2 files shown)


LanguageExt.Core/ClassInstances/Monad/MReaderWriterState.cs
```diff
@@ -0,0 +1,158 @@
+using System;
```
louthy: I think we've gained some local changes that probably shouldn't have been committed?

gwintering:
Whoops, sorry about that.  Fixed.

louthy:
@gwintering  Good stuff, thanks Greg 👍

PR #357, opened 2018-02-08
fix typos? (And probably should call And)

StefanBertels:
I found two issues, **probably** some copy/paste error.


louthy:
Gah!  Thanks 👍

Issue #355, opened 2018-02-07
Record<RECORDTYPE> comparison operators doesn't use CompareTo override


faeriedust:
I made a class public class `VersionNumber : Record<VersionNumber>` and overrode `CompareTo` and called the following code, and it didn't call my ovveride?

`var a = new VersionNumber();
var b = new VersionNumber();
var c = a < b;`

I also couldn't tell from the code with ease or the documentation whether it was supposed to call my override or not, but I feel it's inconsistent that the CompareTo method is virtual and the comparison operators are defined but don't use the CompareTo method.

louthy:
`CompareTo` is implemented automatically.  Why are you overriding it?

bender2k14:
In defense of @faeriedust, why is `CompareTo` declared as `virtual` if you (@louthy) don't expect anyone to override it?  Is the presence of this key word a mistake?

(`Equals` and `GetObjectData` are also declared as `virtual`.  Is the presence of this key words a mistake?)

louthy:
You can override it if you like.  But you’re not getting any functionality for free.  That’s not the problem of the Record type, it’s the problem of the programmer who wants bespoke behaviour over what the Record type provides.

I am not prohibiting overriding because it may be useful in certain circumstances,  but if you want operators you must implement them.

faeriedust:
Can we at least make it more clear via documentation what the implementations for the operators are doing then, if not calling .CompareTo?

There was no documentation, so I stared at the code for like 10 min before giving up, lol.

Is there a reason they can't call the .CompareTo method? Isn't calling .CompareTo normally the expected thing to do for those operators?

bender2k14:
In your PR, what `x` is `null`?

faeriedust:
I was drunkenly and thought that it couldn't be null before, i'll fix it, my bad

faeriedust:
I was drunkenly is the short answer.
I see there are some tests for that case as well, I just had to run them.

`TestClass x = new TestClass(1, "Hello", Guid.Empty);
TestClass y = null;
TestClass z = null;

Assert.True(x > y);
Assert.True(x >= y);
Assert.True(y < x);
Assert.True(y <= x);
Assert.True(y == z);
Assert.True(z == y);`

https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/nullable-types/using-nullable-types

faeriedust:
I was drunkenly is the short answer.
I see there are some tests for that case as well, I just had to run them.

[...]

PR #360, opened 2018-02-09
Change Record operators to call .CompareTo override

faeriedust:
#355


louthy:
Declined.  This is changing the intention of the Record type to support equality of origin only.  I'm not going to stop (and can't stop) developers inheriting from types derived from `Record`, but the equality and ordering operators and methods are not going to support it.  This also brings in performance and memory allocation overheads which is unnecessary when the intention is clear.

PR #365, opened 2018-02-09
Bug fix about order of IgnoreBaseAttribute

### LINKED ISSUE #359: Order of IgnoreBaseAttribute
[This code](https://github.com/louthy/language-ext/blob/async-refactor/LanguageExt.Core/DataTypes/Record/RecordType.cs#L18) (new in `async-refactor`) only considers the first attribute.  If the `IgnoreBaseAttribute` wasn't the only attribute and it didn't come first, then it equivalent to not using it at all.

Can I change to code to consider all attributes instead of just the first?

bender2k14:
Fixes #359


louthy:
Thanks 👍

Issue #374, opened 2018-02-22
Enumerable extension method `Map` called multiple times 


cage433:
The implementation of `Map` calls `Enumerable.Select`, which for `Iterators` results in the lambda being invoked afresh each time we iterate over the result. For example the following fails

```
[Fact]
public void EnumerableMapIsCalculatedOnceTest() {
    var n = 1;

    int NextValue() {
        return n++;
    }

    var actual = Enumerable.Range(0, 5).Map(_ => NextValue());
    var expected = List(1, 2, 3, 4, 5);
    Assert.True(Enumerable.SequenceEqual(actual, expected), $"Expected {toString(expected)} but was {toString(actual)}");
            
    // ensure the result of Map isn't recalculated each time it's iterated over
    Assert.True(Enumerable.SequenceEqual(actual, expected), $"Expected {toString(expected)} but was {toString(actual)}");
}
        
```

bender2k14:
Another way to make this test pass is to change
```
var actual = Enumerable.Range(0, 5).Map(_ => NextValue());
```

to
```
var actual = Enumerable.Range(0, 5).Map(_ => NextValue()).ToList();
```

Do you prefer the solution in your PR more than what I have suggested?  If so, why?

cage433:
Hi

I can see that you're being consistent with `Enumerable.Select`, although
I'm not certain that's a good thing.

Enumerable.Range(0, 5) is fixed, no matter how often I compare it with
List(0, 1, 2, 3, 4) it will be identical. Given that it came as a surprise
that the result of mapping over this range wasn't also constant. Clearly
your suggestion would work, but I'm not keen on having to remember to add a
`ToList` each time I map over an enumerable.

Like yourself I have a massive bias in favour of immutablity, so I would
change your extension method IEnumerable.Map to return a Lst<T>, rather
than IEnumerable<T> - which as we see can change each time it's iterated
over. If I really want the current behaviour I'd just call `Select`.

But that's just me - you've written a cracking library, so no hard feelings
if you disagree with this very new to C# coder.

btw - I think there may be issues with your `Task` extension methods - I've
seen both `Map` and `Bind` block (not every time). In particular you don't
seem to be using `Task.Unwrap` in the latter case. When I get an hour I'll
see if I can come up with a minimal failing test and a PR


cheers,

Alex

bender2k14:
> btw - I think there may be issues with your `Task` extension methods - I've
seen both `Map` and `Bind` block (not every time). In particular you don't
[...]

PR #375, opened 2018-02-22
IEnumerable.Map invokes lambda exactly once for each element

cage433:
Was broken for Iterators.
Fixes issue #374


louthy:
This is braking the lazy nature of IEnumerable.  Seq has the properties you’re after.

cage433:
Thanks - I've limited experience with C#. I'll close the associated pull request

Issue #328, opened 2017-12-26
Issues with use methods


bender2k14:
Does this library have a preferred way to behave when given `null`?

For example, consider the inconsistent behavior of these two `use` overloads from `Prelude_Use.cs`.  If the disposable is obtained form a `Task<>`, then we guard the disposable in case it was `null` but not so if the disposable comes from a `Func<>`.

```
public static async Task<U> use<T, U>(Task<T> computation, Func<T, U> map) where T : class, IDisposable
{
    T t = null;
    try
    {
        t = await computation;
        return map(t);
    }
    finally
    {
        t?.Dispose();
    }
}
```

```
public static R use<T, R>(Func<T> generator, Func<T, R> f)
    where T : class, IDisposable
{
    var value = generator();
    try
    {
        return f(value);
    }
    finally
    {
        value.Dispose();
    }
}
```

louthy:
The second one is a bug.  There’s not a one size fits all policy, but in this case it’s reasonable for use to return null and so it should be protected against in the finally clause.

Data types in general don’t allow null values unless they have the Unsafe suffix.

bender2k14:
Found another bug in the family of `use` methods.  This one is in the extension method in `TryExtensions` with signature `Try<U> Use<T, U>(this Try<T> self, Func<T, U> select)`.  The code `self().Value` should be assigned to `t`.
```
public static Try<U> Use<T, U>(this Try<T> self, Func<T, U> select)
    where T : IDisposable => () =>
    {
        var t = default(T);
        try
        {
            return select(self().Value);
        }
        finally
        {
            t?.Dispose();
        }
    };
```

I would like to repurpose this issue (#328) to address these bugs within the family of `use` methods.

Is that ok with you @louthy or would you prefer if I create separate issues?

louthy:
Fine with me 👍

bender2k14:
There are `use` and `tryuse` methods in `Prelude_Use.cs` as well as some `Use` extension methods.  For these methods, either the `IDisposable` is directly passed in or it is indirectly passed in via one of these monad types:
1. `Func<>`
2. `Task<>`
3. `Try<>`
4. `TryOption<>`
5. `TryAsync<>`
6. `TryOptionAsync<>`

Is there a reason that these monad types have some kind of `use` method while others (such as `Option<>`) don't?

louthy:
> Is there a reason that these monad types have some kind of use method while others (such as Option<>) don't?

[...]

PR #391, opened 2018-04-04
Fix use bugs 328

bender2k14:

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,182 @@
+root = true
+
+[*.cs]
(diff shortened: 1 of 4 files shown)


kdblocher:
@louthy why the reticence in hotfixing your current release? these are common adjustments that have caused quite a few hard-to-notice production issues in our code (other than using post-compilation or analyzers, it is difficult to code around runtime bugs like this)

louthy:
@kdblocher I don't have infinite time and I won't be bullied into pushing code out the door.  This commit has `.editorconfig` changes in which I have no yet had time to review.  So, I'll get to it when I have time.

Issue #377, opened 2018-03-04
Unneeded code following Task.WhenAll


bender2k14:
Recall issue #367.  The [method in question](https://github.com/louthy/language-ext/blob/master/LanguageExt.Core/ClassInstances/Monad/MTask.cs#L197) looks like
```
public async Task<A> Apply(Func<A, A, A> f, Task<A> fa, Task<A> fb) 
{
    await Task.WhenAll(fa, fb);
    return !fa.IsFaulted && !fb.IsFaulted
        ? f(fa.Result, fb.Result)
        : throw fa.Exception;
}
```

I said that this method might be throwing the wrong exception.  Actually, I was wrong.

The [documentation of `Task.WhenAll`](https://msdn.microsoft.com/en-us/library/hh160374(v=vs.110).aspx) says

> If any of the supplied tasks completes in a faulted state, the returned task will also complete in a TaskStatus.Faulted state, where its exceptions will contain the aggregation of the set of unwrapped exceptions from each of the supplied tasks.

and the [documentation of await](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/await) says

> If you await a task-returning async method that causes an exception, the await operator rethrows the exception.
...
> A single task that is in a faulted state can reflect multiple exceptions. For example, the task might be the result of a call to Task.WhenAll.  When you await such a task, the await operation rethrows only one of the exceptions. However, you can't predict which of the exceptions is rethrown.

Therefore, the above method contains unneeded code.  It is equivalent to this simpler method
```
public async Task<A> Apply(Func<A, A, A> f, Task<A> fa, Task<A> fb) 
{
    await Task.WhenAll(fa, fb);
    return f(fa.Result, fb.Result);
}
```

The case is similar for canceled tasks.  The same `Task.WhenAll` documentation says

> If none of the supplied tasks faulted but at least one of them was canceled, the returned task will end in the TaskStatus.Canceled state.

and the same `await` documentation says

> If you await a task-returning async method that's canceled, the await operator rethrows an OperationCanceledException.

[Here](https://github.com/louthy/language-ext/blob/async-refactor/LanguageExt.Core/ClassInstances/Monad/MEitherAsync.cs#L33) is an example in the `async-refactor` branch that includes a call to `Task.IsCanceled` after `await`ing the same task (via a call to `Task.WhenAll`).

[...]

PR #392, opened 2018-04-04
Remove dead task inspection code

bender2k14:
As discussed in #377, removes the dead `Task` inspection code in cases where an exception was going to be thrown anyway.  I am not completely sure about the desired behavior for the remaining cases.  I will ask about them in #377.

### DIFF
--- a/LanguageExt.Core/ClassInstances/Monad/MTaskFirst.cs
+++ b/LanguageExt.Core/ClassInstances/Monad/MTaskFirst.cs
@@ -149,10 +149,7 @@ namespace LanguageExt.ClassInstances
         public async Task<A> Apply(Func<A, A, A> f, Task<A> fa, Task<A> fb) 
         {
             await Task.WhenAll(fa, fb);
-            return fa.IsFaulted && fb.IsFaulted ? throw new AggregateException(fa.Exception, fb.Exception)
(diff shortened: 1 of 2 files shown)


LanguageExt.Core/ClassInstances/Monad/MTaskFirst.cs
```diff
@@ -149,10 +149,7 @@ public struct MTaskFirst<A> :
         public async Task<A> Apply(Func<A, A, A> f, Task<A> fa, Task<A> fb) 
         {
             await Task.WhenAll(fa, fb);
-            return fa.IsFaulted && fb.IsFaulted ? throw new AggregateException(fa.Exception, fb.Exception)
-                 : fa.IsFaulted                 ? throw new InnerException(fa.Exception)
-                 : fb.IsFaulted                 ? throw new InnerException(fb.Exception)
-                 : f(fa.Result, fb.Result);
+            f(fa.Result, fb.Result);
```
louthy: Missing `return`. It's probably worth making sure your code compiles before committing ;)

bender2k14: Oh, sorry *sheepish grin*

Issue #317, opened 2017-12-18
Deadlock from calling Task.Result


bender2k14:
# Example Code
Here is a test that exhibits a deadlock.
```
[Fact]
public void MTaskFold_InAsyncContextWithTaskWaitingForActivation_Halts() =>
  AsyncContext.Run(() => MTaskFold_WithTaskWaitingForActivation_Halts());

private static void MTaskFold_WithTaskWaitingForActivation_Halts() {
  var intTask = TimeSpan
    .FromMilliseconds(100)
    .Apply(Task.Delay)
    .ContinueWith(_ => 0);

  var actual = default(MTask<int>)
    .Fold(intTask, 0, (x, y) => 0)
    (Unit.Default);

  // execution terminates by reaching here
}
```
The type `AsyncContext` is from the NuGet package [Nito.AsyncEx](https://github.com/StephenCleary/AsyncEx).

# Expected Behavior
As stated in the test, the expected behavior is that the test terminates.

# Actual Behavior
The test does not terminate.  Here is the source code under test.
```
[Pure]
public Func<Unit, S> Fold<S>(Task<A> ma, S state, Func<S, A, S> f) => _ =>
{
  if (ma.IsFaulted) return state;
  return f(state, ma.Result);
};
```

The test does not terminate because of a deadlock that is caused when calling `ma.Result`.

# Possible Fix
Stephen Cleary, an expert on this topic and author of the aforementioned NuGet package Nito.AsyncEx, [explains](http://blog.stephencleary.com/2012/07/dont-block-on-async-code.html) why this deadlock occurs.  To summarize, my understanding is that it is a best practice to not call `Task.Result` unless `Task.Status == TaskStatus.RanToCompletion`.  This is _not_ true in the example that I have provided.  Instead it is in the state `TaskStatus.WaitingForActivation`.

Otherwise, the best practice is to `await` the task after calling `Task.ConfigureAwait(false)`.  I don't see how you can do this without changing the return type of `Fold`.  If I may adjust the return type slightly from `Func<Unit, S>` to `Func<Unit, Task<S>>`, then I would implement Stephen's best practice suggestion like this.
```
[Pure]
public Func<Unit, Task<S>> Fold<S>(Task<A> ma, S state, Func<S, A, S> f) => async _ =>
{
    if (ma.IsFaulted) return state;
    return f(state, await ma.ConfigureAwait(false));
};
```

[...]

PR #402, opened 2018-04-13
Fix use bugs 328

bender2k14:

Same as PR #391 except removed the accidental inclusion of `.editorconfig` changes (sorry!).

@louthy, if you have any concerns with one of my PRs, please feel free to let me know and/or request that I change something.  I definitely want to improve LanguageExt by helping you exchange your time for mine :D


bender2k14:
I agree with everything you said.  I think my problem is in correctly understanding the costs.

From previous conversations with you, I know you like to [reduce unnecessary allocations](https://github.com/louthy/language-ext/issues/317#issuecomment-354172338) (i.e. object creations), such as lifting to a `Task<>` (in order to minimize the number of cases to implement).  On the other hand, the impression that I have acquired from reading your code is that you have no qualms with making many [method](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/methods) calls.  My understanding for that position is that the compiler will typically inline these calls.

It is from this understanding that I thought the compiler would inline the lambda in `computation.Map(d => use(d, map))` instead of making an allocation for it.  It was not my intention to increase the number of allocations.

Is there a flaw in my understanding of this?  Is it perhaps the case that the compiler will never inline a lambda expression?

louthy:
> It is from this understanding that I thought the compiler would inline the lambda 

It certainly does turn it into a _method_ in an anonymous class , but it still has allocation costs:

* First class functions and Method Groups are not the same thing in C# (annoyingly).  And so all methods that take a `Func` or a `Action` will have to `new` a `delegate` before use.
* Lambdas that capture free values (arguments that aren't explicitly provided to it) need a class declaration with the free values as member fields.  This obviously needs allocating.

In much of this library it's not a big deal, there's an expected churn of short-lived objects, that probably don't hurt the GC too much because they're in `gen0` and don't ever get promoted to `gen1`.  But for certain functions (like `use`) that are supposed to be more like language features, I think it's much better to consider the impact of all allocations and anything that would be surprising to the end user. 

Where possible it's best to stick to the core grammar of C# (`if`, `else`, `for`, etc.), avoid LINQ, avoid lamdas, avoid heap allocated types.  

I'm sure you will have noticed that a lot functionality that uses the core data-types (like `Option`) will refer directly to their internal members rather than using `Match` to get at the values - it's explicitly for the reason of trying to minimise the overhead of the core functionality so that users of the library get as much for free as possible.  

It's not always the case that I have taken the most optimal route.  It's often for one of the following reasons:
* Premature optimisation is often bad - so it may have been an area that I thought would change
* Optimised code isn't always easy to maintain - so I may have thought the optimal version was unclear
* Large refactors - Sometimes if I'm doing a bulk refactor (like the async refactor) then I'll stick with the safer code but slower code until I get a chance to go back an consider the performance impacts.

It's all pragmatism, obviously.  Just don't look at ugly imperative code and think "that definitely should be more functional", because it's this library's job to deal with that ugliness :)

> Btw, if you're ever unsure, just drop the compiled DLL into ILSpy - you'll see how the inner workings all language features are compiled.

louthy:
Just as a follow up, take this simple class:
```c#
    static class TestClass
    {
        public static void TestMethod(int x)
        {
            var result = Run(() => StaticFunction(x));
        }

        public static A Run<A>(Func<A> f) =>
            f();

        public static int StaticFunction(int a) =>
            a * a;
    }
```
And this is the IL for `TestMethod`:
```IL
.method public hidebysig static 
	void TestMethod (
		int32 x
	) cil managed 
{
	// Method begins at RVA 0x205a
	// Code size 30 (0x1e)
	.maxstack 8

	IL_0000: newobj instance void ConsoleApp5.TestClass/'<>c__DisplayClass0_0'::.ctor()
	IL_0005: dup
	IL_0006: ldarg.0
	IL_0007: stfld int32 ConsoleApp5.TestClass/'<>c__DisplayClass0_0'::x
	IL_000c: ldftn instance int32 ConsoleApp5.TestClass/'<>c__DisplayClass0_0'::'<TestMethod>b__0'()
	IL_0012: newobj instance void class [mscorlib]System.Func`1<int32>::.ctor(object, native int)
	IL_0017: call !!0 ConsoleApp5.TestClass::Run<int32>(class [mscorlib]System.Func`1<!!0>)
	IL_001c: pop
	IL_001d: ret
}
```
At `IL_0000` it calls `newobj` to instantiate the anonymous type `<>c__DisplayClass0_0` (which is allocated on the heap).

It then duplicates the object on the stack, then gets `x` (`ldarg.0`) and stores (`stfld`) it in the `x` field of the  `<>c__DisplayClass0_0` type (which pops one of the copies of the anonymous class's references off the stack).  

It then loads a reference to the `<TestMethod>b__0` method on the stack (which is the lambda function).  After that it creates a new delegate (`Func`) from that reference (which is allocated on the heap).

Finally it invokes it (`call`). 

So, that's two heap allocations.

This is the IL for the anonymous lambda class:
```IL
.class nested private auto ansi sealed beforefieldinit '<>c__DisplayClass0_0'
	extends [mscorlib]System.Object
{
	.custom instance void [mscorlib]System.Runtime.CompilerServices.CompilerGeneratedAttribute::.ctor() = (
		01 00 00 00
	)
	// Fields
	.field public int32 x

	// Methods
	.method public hidebysig specialname rtspecialname 
		instance void .ctor () cil managed 
	{
		// Method begins at RVA 0x2052
		// Code size 7 (0x7)
		.maxstack 8

		IL_0000: ldarg.0
		IL_0001: call instance void [mscorlib]System.Object::.ctor()
		IL_0006: ret
	} // end of method '<>c__DisplayClass0_0'::.ctor

	.method assembly hidebysig 
		instance int32 '<TestMethod>b__0' () cil managed 
	{
		// Method begins at RVA 0x2086
		// Code size 12 (0xc)
		.maxstack 8

		IL_0000: ldarg.0
		IL_0001: ldfld int32 ConsoleApp5.TestClass/'<>c__DisplayClass0_0'::x
		IL_0006: call int32 ConsoleApp5.TestClass::StaticFunction(int32)
		IL_000b: ret
	} // end of method '<>c__DisplayClass0_0'::'<TestMethod>b__0'
}
```
You should hopefully see it's just a regular class, with a `int x` field, a constructor, and a method (which is the lambda's inner code.

louthy:
Now, take a look at this example where I don't capture any free values and everything is `static`:

```c#
    static class TestClass
    {
        public static void TestMethod(int x)
        {
            var result = Run(StaticFunction);
        }

        public static A Run<A>(Func<A> f) =>
            f();

        public static int StaticFunction() =>
            1;
    }
```
And now the IL for `TestMethod`
```c#
.method public hidebysig static 
	void TestMethod (
		int32 x
	) cil managed 
{
	// Method begins at RVA 0x205a
	// Code size 19 (0x13)
	.maxstack 8

	IL_0000: ldnull
	IL_0001: ldftn int32 ConsoleApp5.TestClass::StaticFunction()
	IL_0007: newobj instance void class [mscorlib]System.Func`1<int32>::.ctor(object, native int)
	IL_000c: call !!0 ConsoleApp5.TestClass::Run<int32>(class [mscorlib]System.Func`1<!!0>)
	IL_0011: pop
	IL_0012: ret
}
```
`TestMethod` still needs to allocate a `delegate` on the heap to invoke the `StaticFunction`.  

It's the same for local functions that do and don't capture free values.  It's a shit state of affairs, why the https://github.com/dotnet/csharplang team didn't re-write them as static functions with additional parameters, I don't know, but that's where we are.

bender2k14:
Thank you very  much for that detailed reply!  It was very informative :)

bender2k14:
I believe that I addressed all your comments @louthy in the commit with message [removed delegate allocations](https://github.com/louthy/language-ext/pull/402/commits/4b01ec40908b522ed31e20502fc014dae9072de1).

LanguageExt.Core/DataTypes/Try/Try.Extensions.cs
```diff
@@ -528,13 +526,13 @@ public static Seq<A> AsEnumerable<A>(this Try<A> self)
         this Try<A> self,
         Func<A, Try<B>> bind,
         Func<A, B, C> project) =>
-            MTry<A>.Inst.Bind<MTry<C>, Try<C>, C>(self,    a =>
-            MTry<B>.Inst.Bind<MTry<C>, Try<C>, C>(bind(a), b =>
-            MTry<C>.Inst.Return(project(a, b))));
+            MTry<A>.Inst.Bind<MTry<C>, Try<C>, C>(self, a =>
+         MTry<B>.Inst.Bind<MTry<C>, Try<C>, C>(bind(a), b =>
```
louthy: There appears to be some inconsistent indentation here.  Why aren't they lined up any more?

bender2k14: I didn't change this method.  I think we are seeing changes that entered by branch when I merged in `master` in order to resolve the conflicts.  The green lines are what `master` looks like now.

bender2k14: To address your actual question, the formatting is now partially specified by the `.editorconfig` file.  When auto formatting, sometimes extra white space is allowed to stay.  Other times, extra white space is removed.

Old line 531 begins the method's expression body.  This is a case where extra white space is allowed to stay.  If you move old line 531 to end of line old line 530 and hit "Enter", you will see that Visual Studio would prefer to indent old line 531 exactly as far as old line 530.

In contrast, towards the end of old line 531 and the beginnings of old lines 532 and 533, extra white space was not allowed to stay because the context is now arguments to a method call.  This is one case in which extra white space is not allowed to say.

PR #411, opened 2018-04-28
Fix use bugs 328 retry

bender2k14:

There was PR #402 with requested changes.  I made those changes and pushed them to the branch, but that didn't cause the PR to update (as I expected and [seems is supposed to happen](https://stackoverflow.com/a/9790530/741786)).  Will close that PR.

Now we have this PR instead (*sigh*).


bender2k14:
Oh, figured it out :D.  Closing this PR instead.

PR #418, opened 2018-05-15
changes validation disjunction so it behaves naturally 

StefanBertels:
When validations are combined with disjunction operator they should behave like either data types (or options). That means they should be logically combined like their own "right"/"not right" or "true"/"false" state.

To combine validations that both should hold (AND),  use `&`.
To allow success if at least one validation holds (OR),  use `|`.

Implementation note:
I didn't change FoldValidation.Append. Maybe that should be changed to (other types have Append == Plus == OR).


louthy:
This is wrong.  The || is there to propagate the errors, && is the same as from which doesn’t

PR #424, opened 2018-05-21
Fix 423 bug use disposes too early

### LINKED ISSUE #423: some use overloads disposing too early
We ran into a bug recently because an overload of `use` was disposing too early.  The computation that used the `IDisposable` was asynchronous but the computation the provided the `IDisposable` was not.  This caused the disposal to happen synchronously.

Specifically, calls to `use` were being resolved to the overloads
```
B use<A, B>(A disposable, Func<A, B> f)
```

and
```
B use<A, B>(Func<A> generator, Func<A, B> f)
```

when they needed to be resolved to the (currently nonexistent) overloads
```
Task<B> use<A, B>(A disposable, Func<A, Task<B>> asyncMap)
```

and
```
Task<B> use<A, B>(Func<A> generator, Func<A, Task<B>> asyncMap)
```

I will create a pull request shortly to fix this issue.

bender2k14:
Fixes #423


louthy:
Thanks 👍

PR #425, opened 2018-05-21
UoM Additions

thehoglet:
Force, mass, pressure, temperature.


LanguageExt.Core/DataTypes/UnitsOfMeasure/ScaleTemp.cs
```diff
@@ -0,0 +1,170 @@
...
+            !lhs.Equals(rhs);
+
+        public static bool operator !=(ScaleTemp lhs, AbsTemp rhs) =>
+            !lhs.Equals(rhs);
+
+        public static bool operator >(ScaleTemp lhs, ScaleTemp rhs) =>
+            lhs.Value > rhs.Value;
+
+        public static bool operator >(ScaleTemp lhs, AbsTemp rhs) =>
+            lhs.Celsius > rhs.Celsius;
```
IDisposable: This is inconsistent with the AbsTemp implementation, we should be doing `lhs.Kelvin > rhs.Kelvin` to make sure any precision loss is identical. Same with the other `operator` methods mixing `ScaleTemp` and `AbsTemp`

LanguageExt.Core/DataTypes/UnitsOfMeasure/ScaleTemp.cs
```diff
@@ -0,0 +1,170 @@
...
+
+        public static bool operator <(ScaleTemp lhs, ScaleTemp rhs) =>
+            lhs.Value < rhs.Value;
+
+        public static bool operator <(ScaleTemp lhs, AbsTemp rhs) =>
+            lhs.Celsius < rhs.Celsius;
```
IDisposable: Use `.Kelvin`

LanguageExt.Core/DataTypes/UnitsOfMeasure/ScaleTemp.cs
```diff
@@ -0,0 +1,170 @@
...
+
+        public static bool operator >=(ScaleTemp lhs, ScaleTemp rhs) =>
+            lhs.Value >= rhs.Value;
+
+        public static bool operator >=(ScaleTemp lhs, AbsTemp rhs) =>
+            lhs.Celsius >= rhs.Celsius;
```
IDisposable: Use `.Kelvin`

LanguageExt.Core/DataTypes/UnitsOfMeasure/ScaleTemp.cs
```diff
@@ -0,0 +1,170 @@
...
+
+        public static bool operator <=(ScaleTemp lhs, ScaleTemp rhs) =>
+            lhs.Value <= rhs.Value;
+
+        public static bool operator <=(ScaleTemp lhs, AbsTemp rhs) =>
+            lhs.Celsius <= rhs.Celsius;
```
IDisposable: Use `.Kelvin`

IDisposable:
Why are some `Add` and some `Append` for these?

bender2k14:
Why do you have `AbsTemp` and `ScaleTemp` instead of just one type called `Temperature`?

thehoglet:
@bender2k14 Why ```ScaleTemp``` in addition to ```AbsTemp```? A single ```Temperature``` type is desirable to me, but what is the correct behaviour here:

```
Temperature oneHundredCelsius = 100*degC;
Temperature oneHundredCelsiusinK = 373.15*K;
Print(oneHundredCelsius == oneHundredCelsiusinK); // true

Temperature twoHundredCelsius = 2 * oneHundredCelsius;
Print(twoHundredCelsius.Celsius); // Prints something like 846.3, not 200.0 ???
```

The oneHundredCelsius instance has "forgotten" that it is a Celsius value, but the programmer has not. Is there a way around this? Provide an an alternative API to the * and / operators, [as implemented here](https://github.com/angularsen/UnitsNet/blob/master/UnitsNet/CustomCode/Quantities/Temperature.extra.cs#L89) for instance?

thehoglet:
If the lib goes with separate types for absolute and relative temperature, each is currently missing a converter, I think: ```ScaleTemp ToScale();``` and ```AbsTemp ToAbsolute();``` respectively.

bender2k14:
Oh, I understand now.  Thanks for your explanation.

The fundamental issue is that Kelvin and Celsius have different zero points.

> The `oneHundredCelsius` instance has "forgotten" that it is a Celsius value, but the programmer has not.

Then I think we should have separate types as you have done.  Instead of calling them `AbsTemp` and `ScaleTemp` though, I recommend calling them `Kelvin` and `Celsius`.

louthy:
Ok, my calcs were wrong (fixed now), but you get my point ;)

bender2k14:
> Then only the non-`K` units multipliers like `degC` would ever construct a Temperature with a `ScaleType` that's not `ScaleType.Kelvin`, all other constructions would be with the converted values.

Do you realize that even adding two `Temperature` instances is troublesome when the zero points could be different?  I am unable to tell if you realize that.

louthy:
> Do you realize that even adding two Temperature instances is troublesome when the zero points could be different? I am unable to tell if you realize that.

Maybe I'm missing something, it's been a long day, but if they've both been converted to a common unit then what's the issue?  Obviously there's potential rounding issues of double precision floating point values, but are you referring to something else?

louthy:
Never mind, it's clear now.  Although it's too late for me to think of a solution that isn't ugly.  I'll have a think tomorrow.

bender2k14:
> Obviously there's potential rounding issues of double precision floating point values

Sure.  I am not (yet) considering any edge cases like that though.

> I agree there should be a single Temperature type, not multiple ones.

Here is how I see the problem with with a single temperature type that "knows" what scale type it is and allows two such types to be added together.

The type signatures
```public static Temperature operator +(Temperature lhs, Temperature rhs)```

and
```public Temperature Add(Temperature rhs)```

(and the fact that `Temperature` is a `struct` with a valid default state) ensure that each of these functions form a [magma](https://en.wikipedia.org/wiki/Magma_(algebra)).  I believe it is also true these algebraic structures form a [quasigroup](https://en.wikipedia.org/wiki/Quasigroup).  However, I don't see any way to implement these functions so that their corresponding algebraic strictures have an identity element or are associative.  Maybe there are such implementations though, and I am just failing to see them.

One solution is to have multiple temperature types with explicit and/or implicit conversions between them.

Another solution that is [used here](https://github.com/angularsen/UnitsNet/blob/master/UnitsNet/CustomCode/Quantities/Temperature.extra.cs) (as @thehoglet previously linked to) is to forbid the addition of two instances of type `Temperature`.  Instead, we introduce an additional type that represents a change in temperature.  This is also how the C# standard libraries handle absolute points in time.  The type `DateTime` represents a point in time, and you cannot add two `DateTime`s together, but you can add a `DateTime` and a `TimeSpan` together and get another `DateTime`.

Now after writing all this, I think I like the second solution better.  However, I still don't like the idea of being able to multiple a `Temperature` by a scalar.  Is there really a need for this?  It seems to make about as much sense to multiply a `DateTime` by a scalar, which `DateTime` does not support.

louthy:
I still think the whole thing can be contained within a single `Temperature` type, but for addition and subtraction we only need to know the offset from the zero point and the units that the type is in internally.  So for example if you say `1K + 1°C` then one degree is the same in Kelvin and Celsius, so the result can be `Temperature(ScaleType.Kelvin, 1+1)` for `1K + 1°F` we could do `Temperature(ScaleType.Kelvin, 1+(1*5/9))`.  The left hand side of the operation should dictate the resulting internal units type.

As long as each operator takes into account the units its working with and the operation its performing then I think we can maintain a consistent units system (i.e. we don't have a type per unit, but a type per thing being measured).

Obviously if there was a never ending list of temperature unit types then this wouldn't be a tenable approach.  But there aren't, so I think hiding the messiness for a consistent interface is preferable.

bender2k14:
> > However, I still don't like the idea of being able to multiply a Temperature by a scalar.
>
> Yes, it makes total sense. 10°c * 2 should be 20°c.

The way I am thinking of modeling temperature is as a one-dimensional vector space.  An single temperature measurement is a point and a temperature delta is vector.  In a vector space, there is not usually a multiplication operator that combines a point and a scalar and returns another point.  That is why I said that I don't like the idea of being able to multiply a temperature by a scalar.  What you are suggesting is equivalent to mapping each point to the vector from the origin to that point, multiplying that vector by the scalar, and then mapping back to the point at the head of that vector.  The ambiguity with this is that the origin is not uniquely defined.

Since temperature is a measure of average kinetic energy, I think the most natural way to define multiplication of a temperature and a scalar is to map from the temperature to its corresponding amount of average kinetic energy, multiply this by the given scalar, then from the resulting amount of average kinetic back to its corresponding temperature.  I think this is equivalent to the previous vector space point multiplication by a scalar if we remove the ambiguity there by always internally representing the temperature in Kelvin.

louthy:
I think you're massively over thinking this.

If I get `10 cm` and add `5 cm` to it, I would expect to get a result of `15 cm`.

If I get `10  °C` and add `5  °C` to it, I would expect to get a result of `15  °C`.

I don't need a new type called `CentimetresDelta` or `CelciusDelta`, the operation being performed gives context to the operands.  

The zero point is irrelevant when the operation is addition and subtraction.  Because the 'base point' isn't the zero point, it's the left-hand side.  In my examples above, it's `10 cm` and `10  °C`.

The most important thing when doing delta operations is that the scale of 1 degree is consistent.  So, with  °C and K the size of 1 degree is the same - so if I add `1 °C` to `1 K` then I would expect the result to be `2 K`, but with °F we need to scale the value to be consistent with  °C and K, and ignore the zero-point altogether.  

I feel this whole thing is going around in circles, so I have done an implementation and committed it.  I'd appreciate all interested parties to take a look (and check my calculations).  I've done a handful of unit tests, but if anyone wants to add more, then I'd be eternally grateful.

https://github.com/louthy/language-ext/blob/master/LanguageExt.Core/DataTypes/UnitsOfMeasure/Temperature.cs

bender2k14:
> I don't need a new type called `CentimetresDelta` or `CelciusDelta`, the operation being performed gives context to the operands.

That is not my suggestion.  Sorry for giving you that impression.

louthy:
> That is not my suggestion. Sorry for giving you that impression.

No problem.

I was just thinking a bit more about it, trying to make sure my point is clear.  And it kinda dawned on my that we're probably discussing edge cases here.  Anybody doing operations on temperatures are likely to always be in the same unit anyway, especially for delta operations.  The only time any of this is likely to impact anyone (in the real world) is when converting for display reasons (I suspect).  

The type supports three properties: `Kelvin`, `Celcius`, and `Fahrenheit` to convert between the various units, and I suspect that's the most anybody will ever see of this really.

bender2k14:
> And it kinda dawned on my that we're probably discussing edge cases here. Anybody doing operations on temperatures are likely to always be in the same unit anyway, especially for delta operations - you'd have to be some kind of nihilist to be mixing units in an important calculation. The only time any of this is likely to impact anyone (in the real world) is when converting for display reasons (I suspect).

I completely agree.  So why not eliminate all of those edge cases by using explicitly different types with explicit conversions between them?  This code could still be abstracted so that we don't have any repeated code.

bender2k14:
I think that is because all the other UoM types implemented in Language Ext have a unique zero point.

louthy:
I have made the addition and subtraction operators use the absolute values now.  If someone is adding `1°C` to `1K` then that's something they'll find in testing when nothing works as they expected, having a single type called `Temperature` is more valuable in my opinion (the only other viable solution in my mind is to change all unit types to be the measurement type rather than the concept it's measuring - which risks breaking a lot of user code).

Anyway, it's not worth losing sleep over and I'd prefer we concentrated on getting the rest of this pull request into a good state.

PR #433, opened 2018-06-06
fix small bug in reduce (was same order as reduceBack)

StefanBertels:
Reduce probably was indented to work forward, ReduceBack reverse.


louthy:
Thanks :)

louthy:
Sorry for the delay in deploying, I thought I'd already merged this.  On nuget now 3.0.16.  Thanks again.

PR #437, opened 2018-06-13
Improve type parameter names in compose extensions

bender2k14:
When a function has multiple type parameters, I have seen two styles for naming them in Language Ext.  One to name them like `T1`, `T2`, `T3` and the other is to name them like `A`, `B`, `C`.  I have come to really prefer the latter.

This PR changes form the first style to the second style for the functions in one file.  It also changes the parameter names (that were `a` and `b`) to avoid confusion with this new type parameter names.


louthy:
Agreed, the A B C route is my preferred approach now. It just feels more
elegant than T, dunno why.

Will check the pull request when I get back to my pc

PR #440, opened 2018-06-15
Fixes typo in Bind parameter type in README.md

blabaere:
In `Monad.Bind` method, changes the type of the `bind` parameter from `Func<B, MB>` to `Func<A, MB>`.


louthy:
Well spotted, thanks!

Issue #445, opened 2018-06-27
TaskExtensions is not in the LanguageExt namespace


mattfurness:
I recently added Langauge Ext to an existing codebase and ran into an issue where it started calling [`Select`](https://github.com/louthy/language-ext/blob/v3.0.5/LanguageExt.Core/DataTypes/Task/Task.Extensions.cs#L55) from `TaskExtensions` instead of an extension method in a namespace that I had explicitly imported because the `TaskExtensions` class is [not in any namespace](https://github.com/louthy/language-ext/blob/v3.0.5/LanguageExt.Core/DataTypes/Task/Task.Extensions.cs#L11).  It was a subtle issue that took a little while to track down because I was not explicitly importing anything from Language Ext.

I was just wondering if not putting `TaskExtensions` in the `LangaugeExt` namespace was intentional, or was just accidentally missed. 

Thanks for all your effort on this library @louthy.

louthy:
@mattfurness  Hi Matt, it was a conscious decision to do this, because:

* Really there's only one way to compose Tasks, which is the way that the `async` keyword does it: continue on success, bail out on fail/cancel.  So replicating that in `Select` and `SelectMany` made sense.
* It also made sense that it should be a feature of `Task` already and not something to explicitly opt in to.  i.e. `Task` should be a `Task` monad, it just made sense.

I can see the issue though, mostly when I've done this in the past (putting the extensions in the global namespace) it has been to extend types that belong to lang-ext, so I can totally see your point that it can cause clashes with other implementations.

Obviously moving it to the `LanguageExt` namespace is what needs to happen.  Although also obviously it sounds like that will still preclude you from using the `LanguageExt` namespace along with your other extensions, but at least you'll have the choice.  

Out of interest what are you doing in your version of `Select` for `Task`?  Is it the same mapping functionality?

mattfurness:
Thank you for the super quick reply @louthy 😄 

Regarding your comment on precluding the use of the namespace:
- Fortunately we don't need to import `LanguageExt` in the file in question, so it wouldn't be an issue in this specific case.
- If I had needed to import the `LanguageExt` namespace and the tests started failing in the way they did it would have been much more obvious. It was the implicitness that caught me out.

Regarding the version of `Select` in the existing project, I'll try to describe it succinctly:
[...]

PR #446, opened 2018-06-29
Move TaskExtensions into the LanguageExt namespace

mattfurness:
This should avoid the implicit ambiguous method call issue raised in #445.

Note: This seems to build fine, I was seeing some of the async tests fail intermittently even before I made any changes, so I am as sure as I can be that I didn't break anything, but it is probably worth you running the tests yourself @louthy


louthy:
Thanks @mattfurness

PR #448, opened 2018-07-02
fix logic bug with biconditional in bigint

StefanBertels:
Seems like a logic bug here. Is this fix correct? If you want I can add this for the other types (same thing with at least TInt, TLong).


louthy:
@StefanBertels Looks like TInt, TLong, and TShort make the same mistakes.  Do you want to fix those too?

louthy:
Thanks @StefanBertels 👍

Issue #450, opened 2018-07-11
TryExtensions.Use passes null to the select function when the constructor call fails


simonmckenzie:
This seems unintuitive to me, but perhaps it's supposed to be this way - when using `Try(...).Use(...)`, if the function passed to `Try` throws an exception, the function passed to `Use` will be supplied with a null parameter, and the exception is effectively lost.

``` csharp
Try<IDisposable>(() => throw new Exception())
.Use(disposable => 
{
    Console.WriteLine(disposable == null ? "null" : "not null");
    return 0;
}).Try();
```
The output from the above code will be "null".

It seems to me that where the underlying implementation simply passes `self().Value` to `Prelude.use`, it should probably only do this if `self()` is not in a faulted state.

From [`Try.Extensions.cs`](https://github.com/louthy/language-ext/blob/master/LanguageExt.Core/DataTypes/Try/Try.Extensions.cs):
``` csharp
[Pure]
public static Try<U> Use<T, U>(this Try<T> self, Func<T, U> select)
    where T : IDisposable => () =>
    use(self().Value, select);
```

Am I correct?

bender2k14:
Yes, I think you are correct. Would you like to make a pull request to fix this?

simonmckenzie:
Sure!

simonmckenzie:
Done. Note that I have observed two tests which fail intermittently, but they're not related to my changes.
```csharp
OptionAsyncTests.Issue206()
MemoTests.MemoTest3()
```

louthy:
> OptionAsyncTests.Issue206()
> MemoTests.MemoTest3()

Yes, there are a few time-critical tests which often fail when running all tests.  I usually run them by hand after doing a full run.  Don't worry too much about those. 

The pull request is accepted, thanks.

PR #451, opened 2018-07-11
Updated the Use extensions to not call use on fail

simonmckenzie:
#450 
The use extensions previously passed `self().Value` directly into `Prelude.use`, which resulted in a null input value to the consuming function when the initial `Try` was in a faulted state


louthy:
Thanks Simon 👍

PR #465, opened 2018-08-19
Fixed XML comments for BiConditions

colethecoder:
It looks like a recent commit put the comments out of sync with the logic for some of the BiCondition functions.


louthy:
Cheers Tom

PR #480, opened 2018-09-18
some tests and fixes around setitem/trysetitem for maps and sets

StefanBertels:
Besides some bugs (see diff) I changed two more things that aren't bugs:

1. I changed some test names and moved one test to another file, hope this is better now.

2. I modified all SetItem/Set/TrySetItem/TrySet variants so the key is replaced always. This makes it more consistent and I hope this does not have issues I didn't see. 
I don't think it is a good idea to rely on this feature in user code (key is defined as equal there), but at least LanguageExt data types have to decide which key to store. Using always the latest key makes it e.g. consistent with Remove(oldKey).Add(newKey, newValue).

I hope I found all issues.


louthy:
Looks good, thanks @StefanBertels

PR #486, opened 2018-09-22
Supplement HeadOrNone with Validation and Either flavours

colethecoder:
HeadOrNone is cool for matching up with Options but there are times when you need to match the Head with Validation or Either and in these cases the code tends to end up as:

```csharp
list
.HeadOrNone()
.ToValidation(....)
```
I've added HeadOrInvalid and HeadOrLeft to save having to go via Option.


louthy:
@colethecoder  Looks good, thanks.  In London for a few days, so will roll out early next week.

Issue #488, opened 2018-09-29
SeqEnumerable<T> prevents proper disposal of the IEnumerator


faeriedust:
It seems that the implementation of `SeqEnumerable<T>` makes it difficult to dispose the `IEnumerator<T>` from the original `IEnumerable<T>`.

I originally ran into this issue when I attempted to call `.HeadOrNone<T>()` on an `IQueryable<T>` before committing a transaction, causing an SqlException ("The transaction operation cannot be performed because there are pending requests working on this transaction.")

Here are a couple tests using NSubsitute that demonstrate the problem:

```
using System.Collections.Generic;
using System.Linq;
using LanguageExt;
using NSubstitute;
using Xunit;

namespace BlocherConsulting.Testing.UnitTests {
  public class EnumerableTests {

    //Passes
    [Fact]
    public void NormalEnumerableAllowsEnumeratorToBeDisposed() {
      //ARRANGE
      var enumerator = Substitute.For<IEnumerator<int>>();
      var enumerable = Substitute.For<IEnumerable<int>>();
      enumerable.GetEnumerator().Returns(enumerator);

      //ACT
      enumerable.Apply(Enumerable.FirstOrDefault);

      //ASSERT
      enumerator.Received().Dispose();
    }

    //Fails
    [Fact]
    public void SeqEnumerableDoesNotAllowEnumeratorToBeDisposed() {
      //ARRANGE
      var enumerator = Substitute.For<IEnumerator<int>>();
      var enumerable = Substitute.For<IEnumerable<int>>();
      enumerable.GetEnumerator().Returns(enumerator);

      //ACT
      enumerable.Apply(Prelude.Seq).Apply(Enumerable.FirstOrDefault);

      //ASSERT
      enumerator.Received().Dispose();
    }
  }
}
```

faeriedust:
I suppose possible fix is to just ensure that the original `IEnumerator<T>` gets disposed when the one from `Seq<T>` is disposed?

faeriedust:
@louthy Your fix doesn't address my issue though. 

It shouldn't matter if I read the last item or not. It's extremely common to not read the last element of an `IEnumerable`, particularly if there isn't one.  If I dispose the `IEnumerator` from the Seq, it means I am done reading and therefore you should release the resources, e.g. dispose the enumerator from the original Seq, no?

I can adjust the tests I have above to include nonempty sets and they will fail even with your fix.
```
//Passes
[Fact]
public void NormalEnumerableBehavior()
{
[...]

PR #489, opened 2018-09-29
Feature/seq enumerable

faeriedust:
Added test and missing using statement to SeqEnumerable2


faeriedust:
Fixes #488

louthy:
That won't fix it, it will dispose after the first item.  I have a fix for it that I'll commit.

PR #493, opened 2018-10-04
fixes iterator disposal issue with .HeadOrNone()

faeriedust:
Fixes the issue with calling `.HeadOrNone()` on an `IQueryable` not disposing the enumerator
#488


faeriedust:
Those two are harder to fix since they call methods that are baked into `Seq`, but I can take a shot at moving them off of it?

louthy:
@faeriedust I have done this and deployed in `v3.0.30`

PR #497, opened 2018-10-08
Implement MapFail and BiMap for ValidationSeq

michael-wolfenden:


louthy:
@michael-wolfenden Thanks for this, looks good, we should update the other validation type: `Validation<MonoidFail, FAIL, SUCCESS>` they're both part of the same story and should have API parity where appropriate.

louthy:
Deployed in `v3.0.30`

PR #502, opened 2018-10-10
BindBottom for Either (was: ifBottom / .IfBottom)

StefanBertels:
Adds ifBottom / .IfBottom for either allowing fluent removal of bottom state (e.g. after .Filter).

I missed this for code like:

```
var aOption = Some(1);
var bEither = Right<string, int>(2);  // or Left<string, int>("no value");

Either<string, int> resultOrMessage = from a in aOption.ToEither("missing a")
    from b in bEither.Filter(isPositive).IfBottom("b is not positive")
    select a + b;
```
Without IfBottom I wouldn't get a left state containing e.g. error information because of Filter.

I implemented only one signature (parameter `Func<Either<A,B>> Bottom`). If you think this is useful, I can add some more cases.

I guess this is most useful with Left<A,B> as fallback value, so we could reduce universality and just allow a left value like this (easier API):

`Either<L,R> IfBottom(Func<L> Bottom)` resp. `Either<L,R> IfBottom(L leftValue)`


louthy:
This is a good idea.  But, I think they're more like functor maps (because they return `Either` rather than a concrete value), maybe this instead:

```c#
    Either<L, R> MapBottom(Func<L> Bottom);
    Either<L, R> MapBottom(Func<R> Bottom);
    Either<L, R> BindBottom(Func<Either<L, R>> Bottom);

    Either<L, R> MapBottom(L Bottom);
    Either<L, R> MapBottom(R Bottom);
    Either<L, R> BindBottom(Either<L, R> Bottom);
```
That's then more consistent with `MapLeft` and `BindLeft`.

Also, there's `EitherUnsafe<L, R>`, and `EitherAsync<L, R>` too.  It's good to keep consistency across the types :)

louthy:
btw, if you're using `Bottom` because of `where` clauses in your `Either` expressions, then you can do something like this for more useful filtering and contextual alternative values:
```c#
    var res = from x in eitherA
              from r in x > 5
                  ? Right<string, int>(x * 2)
                  : Left<string, int>("Number is less than 5")
              select r;
```

StefanBertels:
Hi, 

I changed to BindBottom and will add a commit soon with async support.

I struggle currently with MapBottom because I tried to first build a static function in Choice.Prelude.cs:

```
/// <summary>
        /// Returns the input if in Left or Right state
        /// Returns the result of bottom function if in Bottom state as Left value
        /// </summary>
        /// <param name="bottom">Function to generate a Left value if in the Bottom state</param>
        /// <returns>Returns an either value</returns>
        [Pure]
        public static Either<A, B> mapBottom<CHOICE, CH, A, B>(CH ma, Func<A> bottom)
            where CHOICE : struct, Choice<CH, A, B> =>
            default(CHOICE).Match(ma,
                Left: Left<A, B>,
                Right: Right<A, B>,
                Bottom: bottom.Compose(Left<A, B>));

        /// <summary>
        /// Returns the input if in Left or Right state
        /// Returns Left(bottom) if in Bottom state
        /// </summary>
        /// <param name="bottom">value to return as Left value if in Bottom state</param>
        /// <returns>Returns an either value</returns>
        [Pure]
        public static Either<A, B> mapBottom<CHOICE, CH, A, B>(CH ma, A bottom)
            where CHOICE : struct, Choice<CH, A, B> =>
            default(CHOICE).Match(ma,
                Left: Left<A, B>,
                Right: Right<A, B>,
                Bottom: () => Left<A, B>(bottom));

        /// <summary>
        /// Returns the input if in Left or Right state
        /// Returns the result of bottom function if in Bottom state as Right value
        /// </summary>
        /// <param name="bottom">Function to generate a Right value if in the Bottom state</param>
        /// <returns>Returns an either value</returns>
        [Pure]
        public static Either<A, B> mapBottom<CHOICE, CH, A, B>(CH ma, Func<B> bottom)
            where CHOICE : struct, Choice<CH, A, B> =>
            default(CHOICE).Match(ma,
                Left: Left<A, B>,
                Right: Right<A, B>,
                Bottom: bottom.Compose(Right<A, B>));

        /// <summary>
        /// Returns the input if in Left or Right state
        /// Returns Right(bottom) if in Bottom state
        /// </summary>
        /// <param name="bottom">value to return as Left value if in Bottom state</param>
        /// <returns>Returns an either value</returns>
        [Pure]
        public static Either<A, B> mapBottom<CHOICE, CH, A, B>(CH ma, B bottom)
            where CHOICE : struct, Choice<CH, A, B> =>
            default(CHOICE).Match(ma,
                Left: Left<A, B>,
                Right: Right<A, B>,
                Bottom: () => Right<A, B>(bottom));
                
```

The problem is that I have to change the return type to CH here to make it generic (using Choice). This means I need to create a `Left`. I got this solution for the `Right` path:

```
        [Pure]
        public static CH mapBottom<CHOICE, MONAD, CH, A, B>(CH ma, Func<B> bottom)
            where CHOICE : struct, Choice<CH, A, B>
            where MONAD : struct, Monad<CH, B> =>
            default(CHOICE).IsBottom(ma) ? default(MONAD).Return(bottom()) : ma;
```

Is there any typeclass support for creating left? (Otherwise this variant will have to be non-generic, i.e. without using `Choice`).

StefanBertels:
I think there are three ways to go:

(1) just merge the commits until now -- getting only BindBottom and dropping MapBottom.
(2) implement MapBottom just for Either/EitherAsync (non-generic / without `Choice`)
(3) create an interface like 'IReturn<T>' as the `return` part of the monad definition which could be implemented by Either as `IReturn<L>`. We could even replace the `return` part of the monad by `IReturn<A>` so Option would implement it once and Either would implement it twice (for L and R). Seems like there might already be an interface in LanguageExt that might had this target: `Liftable`. It's currently unused..

When going way (3) we should do this in a separate issue. I'm glad to help with a PR. This option obviously is the most work and doesn't pay off just for MapBottom. But maybe `Liftable` is a useful concept in other parts, as well.

louthy:
> Is there any typeclass support for creating left? (Otherwise this variant will have to be non-generic, i.e. without using Choice).

```c#
    default(MONAD).Fail(bottom())
```

louthy:
https://github.com/louthy/language-ext/issues/911

PR #503, opened 2018-10-11
Add missing Match overload from Seq to IEnumerables

michael-wolfenden:
`Seq`  had an overload of `Match` that allowed

```csharp
Seq(1, 2, 3)
	.Match(
		Empty: () => "",
		Seq: xs => string.Join(", ", xs)
	);
```

This overload was missing for IEnumerables

```csharp
// does not compile
Set(1, 2, 3)
	.Match(
		Empty: () => "",
		More: xs => string.Join(", ", xs)
	);
```


louthy:
Thanks @michael-wolfenden 👍

PR #506, opened 2018-10-17
Seq perf refactor

louthy:


LanguageExt.Core/DataTypes/Seq/Enum.cs
```diff
@@ -0,0 +1,83 @@
...
+
+        ~Enum()
+        {
+            iter?.Dispose();
+            iter = null;
+        }
+
+        public (bool Success, A Value) Get(int index)
+        {
+            if (index < list.Count)
```
IDisposable: Is this safe if we're mid-growing? Will the `list.Count` be partially updated before we're finished and thus this could return uninitialized array element?  In other words, do we need to wrap in a `lock(list)`?

LanguageExt.Core/DataTypes/Seq/Enum.cs
```diff
@@ -0,0 +1,83 @@
...
+            if (iter == null)
+            {
+                return (false, default(A));
+            }
+            else
+            {
+                lock (list)
+                {
+                    bool theresMore = true;
+                    while (index >= list.Count && (theresMore = iter.MoveNext()))
```
IDisposable: Do we want to move the `lock(list)` down here so we unlock between each iteration of `.Add` in case the incoming iterator is really long and you've asked for a really high `index`?

LanguageExt.Core/DataTypes/Seq/Enum.cs
```diff
@@ -0,0 +1,83 @@
...
+                    {
+                        list.Add(iter.Current);
+                    }
+                    if (!theresMore)
+                    {
+                        iter = null;
+                    }
+                }
+            }
+            if (index < list.Count)
```
IDisposable: Since this is common with the earlier code in the get, maybe we should check and grow, then fall through?

LanguageExt.Core/DataTypes/Seq/Enum.cs
```diff
@@ -0,0 +1,83 @@
...
+            {
+                return (true, list[index]);
+            }
+            else
+            {
+                return (false, default(A));
+            }
+        }
+
+        public void Strict()
```
IDisposable: Comment:
So is `Strict` the name of the _early_ memoization?

LanguageExt.Core/DataTypes/Seq/Enum.cs
louthy: `_size` in `List<T>` is always updated after the allocation if its internal array and any set operation.  So, even though it could be out of sync with the backing store, it will only ever report that the collection is smaller than it thinks.  So, the worst that could happen (I believe) is that we attempt to stream another item.   But there shouldn't be any synchronisation issues.

There's a unit test that pretty aggressively tests the lazy loading whilst multiple threads are trying to iterate it:

https://github.com/louthy/language-ext/blob/c412a962fac7110845b80d0f80a467cc929637ff/LanguageExt.Tests/Seq.Module.Tests.cs#L182

LanguageExt.Core/DataTypes/Seq/Enum.cs
louthy: I guess there's one of two trade-offs needs to be picked here:

1. The lazy stream blocks if you try to jump ahead 
2. The general performance if lazy streams is reduced from the cost of the `lock` acquire per iteration

Going to have a think on this one.  And probably do some perf tests to see where we stand.  I'm erring towards your solution, because the type will act more like a regular immutable type.

LanguageExt.Core/DataTypes/Seq/Enum.cs
louthy: Sorry, I don't follow?  I'm looking at it now, and I can see that last `if` should be moved inside the `else` above it.  But other than that not really sure.  The ugliness of the function is obviously to try and return as quickly as possible without doing unnecessary work.

LanguageExt.Core/DataTypes/Seq/Enum.cs
louthy: I guess you could say that.  More precisely `Strict` forces evaluation.

LanguageExt.Core/DataTypes/Seq/Enum.cs
louthy: I have made the change you suggested.

LanguageExt.Core/DataTypes/Seq/Seq.cs
```diff
@@ -808,20 +808,14 @@ public override int GetHashCode()
...
-                 x.GetHashCode() != y.GetHashCode()))
-            {
-                return false;
-            }
-
-            return Enumerable.SequenceEqual(this, rhs);
-        }
+        public bool Equals(Seq<A> rhs) =>
+            !ReferenceEquals(rhs, null) &&
+            this.seq == null && rhs.seq == null &&
```
IDisposable: Is it possible for `ReferenceEquals(this.seq, rhs.seq)` to be true (e.g. could we ever have two different `Seq<>` that have the same inner `seq`)?  If so, would that be worth inserting between lines 812 and 813?

LanguageExt.Core/DataTypes/Seq/Seq.cs
```diff
@@ -808,20 +808,14 @@ public override int GetHashCode()
...
+            this.Count == rhs.Count &&
+            this.GetHashCode() == rhs.GetHashCode() &&
+            Enumerable.SequenceEqual(this, rhs)
+                 ? true
```
IDisposable: Why is this ternary here? Are we just using this to make sure that exceptions or side-effect that might be happen deep in `Enumerable.SequenceEqual` get a chance?

PR #508, opened 2018-10-24
Fixing EitherAsync.Bottom 

StefanBertels:
Internal `data` was null (runtime error when calling IsBottom).

Same fix is included in #502 but that PR might not be ready to merge now.

### DIFF
--- a/LanguageExt.Core/DataTypes/EitherAsync/EitherAsync.cs
+++ b/LanguageExt.Core/DataTypes/EitherAsync/EitherAsync.cs
@@ -31,11 +31,11 @@ namespace LanguageExt
     public struct EitherAsync<L, R> :
         IEitherAsync
     {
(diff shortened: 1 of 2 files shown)


StefanBertels:
`default(EitherAsync)` is now bottom state.

Note: I changed constructor so null argument will throw an exception instead of silently falling back to bottom. IMHO this makes it more clean and one should never pass null to the constructor to get some EitherAsync in bottom state.
If you think the previous behaviour is better (because there might be legitimate ways to indirectly have null value here with bottom "meaning") we can change that back, of course. My thought here was just "another null check is fine".

louthy:
@StefanBertels Thanks for this, sorry for the delay, I've done a slightly [DRYer version](https://github.com/louthy/language-ext/commit/79f3c98b86f5c91a81b7033aecdb909d36fd3f89).  So, going to close this.

PR #512, opened 2018-10-24
OptionAsyncTests.Issue206 fails

StefanBertels:
Running the test OptionAsyncTests.Issue206 standalone worked fine mostly, but running a bunch of test not (probably due to exhausting task scheduler).

The commit e90fdf5 changes the test to improve reproducibility.

Problem is that `IfSome` is called with an async action which will not be awaited because it's only an `Action<T>` then. Replacing the call to `IfSomeAsync` (in test) should solve the issue.

How to avoid this? Could be a (really bad) issue for other async code... :-(

Maybe the best (clean) option is to drop support for `Action<A>` parameters for async types by deleting all async methods accepting `Action<A>`.

This will force to use `Func<A, Unit>` instead. One can still do bad things, but it's more difficult to do.

BTW: I found this on Stackoverflow: https://stackoverflow.com/questions/19024014/check-if-action-is-async-lambda which could be used to at least throw a runtime exception. But I'm not recommending this.

### DIFF
--- a/LanguageExt.Tests/OptionAsyncTests.cs
+++ b/LanguageExt.Tests/OptionAsyncTests.cs
@@ -78,6 +78,7 @@ namespace LanguageExt.Tests
             OptionAsync<int> optAsync = Some(4).ToAsync();
             var tskOptionAsync = optAsync.IfSome(async (i) =>
             {
+                await Task.Delay(100);
                 var x = DoWork();
 
                 lock (sync)
@@ -95,8 +96,8 @@ namespace LanguageExt.Tests
             }
             await tskOptionAsync;
 
-            Assert.True(tasks[0].Status == TaskStatus.RanToCompletion);
-            Assert.True(tasks[1].Status == TaskStatus.RanToCompletion);
+            Assert.Equal(TaskStatus.RanToCompletion, tasks[0].Status);
+            Assert.Equal(TaskStatus.RanToCompletion, tasks[1].Status);
         }
 
         [Fact]


StefanBertels:
Another link: https://stackoverflow.com/questions/41588195/how-to-prevent-an-action-parameter-from-being-an-async-lambda 

My commit https://github.com/louthy/language-ext/pull/512/commits/6648f2e65e061577e37a9da035466da5e213259c fixes the issue by adding a (probably missing) variant `IfSome(Func<A, Task>)` which gets called by the current test code.

The async lambda produces an anonymous function returning `Task` and this can be awaited. `Func<A, Task>` variant gets called on this type.

There is still the problem that it will break if the async lambda gets casted to a simple `Action` somewhere. Probably there are three options:

(1) It's not a bug in the library => user error => don't care (besides implementing Func<A, Task> variants)
(2) remove any `Action<A>` variants to reduce the problem as far as possible
(2b) mark Action variants at least as [Obsolete]
(3) check the action for async on runtime and handle this some way. This is a code smell, for sure. And throwing an exception isn't correct as this just might be some fire-and-forget-action by intent. I'm not sure whether there is a hack to await something like this.

StefanBertels:
Added some comment fixes.

When going way (1) there probably every method accepting `Action` should have a companion with `Func<Task>`. I'm not sure whether this helps making the library better as many methods already are defined with a different name (like `MatchAsync`). 

See https://github.com/louthy/language-ext/blob/267f1c69002a58df64655f777cd8b6fee57f5e7c/LanguageExt.Core/TypeClasses/OptionalAsync/OptionalAsync.cs#L47-L73

There is a `Match` function defined without this companion.

(It's such a mess that C# allows `async void` / passing async functions to `Action`.)

StefanBertels:
(1b): implement all those companion functions (with `Func<Task>` instead of `Action`) and mark all of them obsolete and give a hint to use `MatchAsync` instead of `Match` (and `IfSomeAsync` instead of `IfSome` etc).

PR #513, opened 2018-10-25
Fixed text for hexadecimal

colethecoder:
The text was wrong for hexadecimal in parsec, looks like a copy paste issue from octal


louthy:
Good stuff, cheers Tom

PR #518, opened 2018-11-02
Bug fix517 throw on option map to null

### LINKED ISSUE #517: Unexpected behavior when calling Option.Map to null value.
I stumbled on this by happenchance earlier, but this test passes, which felt very strange to me.

```
[Fact]
public void ThisPasses() {
  object x = null;

  Option<object> result = Some(new object())
    .Map(_ => x);

  Assert.Equal(None, result);
}
```

I was expecting to get a `LanguageExt.ValueIsNullException`.

If I had intended this behavior, then my impression is that I should have called `.Bind`, not `.Map`

bender2k14:
Fixes #517.

I think this is the correct fix.  All the tests, including the newly added one, pass.


louthy:
Thanks!

Issue #530, opened 2018-12-04
Record failure on .NET (ok until 3.0.28)


StefanBertels:
I updated LanguageExt in a project and now get errors.

Version 3.0.28 works, Version 3.0.31 and 3.1.13 raise errors in my project.
I have a unit test showing the problem in my code but I could not reproduce it outside of my project. After copying the test and types into LanguageExt source code (test project) the test is green there (current master == 3.1.13). I'm confused why same test shows different behaviour.

My guess: it is something around 2f1e06. 

```
System.TypeInitializationException: The type initializer for 'LanguageExt.RecordType`1' threw an exception.
   at LanguageExt.Record`1.Equals(RECORDTYPE other)
   at MyProject.Actors.MyStaticClass.MyRecordClass.Equals(MyRecordClass rhs) in MyStaticClass.cs:line 488
   at Xunit.Sdk.AssertEqualityComparer`1.Equals(T x, T y)
   at Xunit.Assert.Equal[T](T expected, T actual, IEqualityComparer`1 comparer)
   at Xunit.Assert.Equal[T](T expected, T actual)
   at MyProject.Actors.MyTests.TestMyRecordClass() in MyTests.cs:line 22
System.InvalidOperationException: CompileToMethod cannot compile constant 'MyProject.Actors.MyStaticClass+MyRecordClass' because it is a non-trivial value, such as a live object. Instead, create an expression tree that can construct this value.
   at System.Linq.Expressions.Compiler.BoundConstants.EmitConstant(LambdaCompiler lc, Object value, Type type)
   at System.Linq.Expressions.Compiler.LambdaCompiler.EmitConstant(Object value, Type type)
   at System.Linq.Expressions.Compiler.LambdaCompiler.EmitConstantExpression(Expression expr)
   at System.Linq.Expressions.Compiler.LambdaCompiler.EmitExpression(Expression node, CompilationFlags flags)
   at System.Linq.Expressions.Compiler.LambdaCompiler.EmitBinaryExpression(Expression expr, CompilationFlags flags)
   at System.Linq.Expressions.Compiler.LambdaCompiler.EmitExpression(Expression node, CompilationFlags flags)
   at System.Linq.Expressions.Compiler.LambdaCompiler.EmitUnliftedAndAlso(BinaryExpression b)
   at System.Linq.Expressions.Compiler.LambdaCompiler.EmitAndAlsoBinaryExpression(Expression expr, CompilationFlags flags)
[...]

PR #531, opened 2018-12-05
Reproduce problem #530 (tests fail)

StefanBertels:
Record type IL code feature fails in .net Framework, e.g. net46 or net462.

This change adds net46 to targetframeworks of the test project which will make some tests fail.
See #530.


louthy:
Thanks!

Issue #221, opened 2017-05-06
Overriding Eq and Ord in NewType, other NewType questions


jmagaram:
I'm trying to get my head around this TypeClass stuff. It seems like you're using generic type parameters to inject/customize behavior versus "the usual way" of perhaps overriding some virtual methods. It seems like this TypeClass approach is the ONLY way to implement static methods, like == operators, without reimplementing those static methods on every derived class. Few questions:

Taking this approach to its logical conclusion shouldn't NewType be `NewType<TValue,TPredicate,TEquals,TOrd>`? If I wanted a wrapped string to have different comparison behavior than the default (like case insensitive), I'd have to override a bunch of methods on the derived NewType class and re-implement some operators. Correct?

Is the reason why you don't define NewType that way because the generic name gets unwieldy to define?

Finally, it really seems like NewType should be a struct for performance reasons. However if it was struct you could never give it a friendly name since there is no inheritance. Is that why you implement this as a class rather than a struct?

louthy:
> Taking this approach to its logical conclusion shouldn't NewType be NewType<TValue,TPredicate,TEquals,TOrd>?

Yes, and at some point I will add those variants too (if you need it soon then I can bring that forward, as a lot of the new-type system has settled down since the v2 release).  As you can probably imagine it ends up being quite a bit of typing to support everything that the new-types, num-types, and float-types can do.  Mostly new-types are simple types with equality and comparison operators 'built in'; but as you point out, types like `string` can have different strategies for equality.

> Finally, it really seems like NewType should be a struct for performance reasons. However if it was struct you could never give it a friendly name since there is no inheritance. Is that why you implement this as a class rather than a struct?

You can't derive from a `struct`, so the new-type system wouldn't work at all.  But yes, I'd much prefer it if C# allowed derived structs, or classes with value-type properties.

jmagaram:
> You can't derive from a struct

There might be a way of using a Roslyn script during build to simulate this. It could take a struct<T,U>, plug in a specific T and U, and output a partial struct file.

StefanBertels:
@louthy Do you think you could/would reopen this feature request?

I think it's quite useful to have som case-insensitive string type: 
```
public class MyType : NewType<MyType, EqStringOrdinalIgnoreCase, string, StrLen<I1, I400>>
{
        public MyType (string value) : base(value)
        {
        }
}
```
or
```
public class MyType : NewType<MyType, EqStringOrdinalIgnoreCase, string>
{
[...]

PR #540, opened 2019-01-02
Add ordering type class to NewType

gwintering:
resolves louthy/language-ext #221


louthy:
A cursory glance and this looks excellent, but I don’t have a huge amount of time to review right now.  Do you have an urgency in getting this released?

gwintering:
No there's no hurry at all.  For my own purposes I've created a subclass of the existing `NewType` that accepts an `Ord<>` and uses it in overrides.  But I wanted to contribute something a little more idiomatic into the library.

louthy:
Looks great, thanks!

PR #541, opened 2019-01-03
Add Regex match string predicate

gwintering:
Review at your leisure


louthy:
Even though I have a fundamental mistrust of regex, I'll allow, thanks

PR #543, opened 2019-01-09
Bugfix for Try / TryOption (mostly bottom state)

StefanBertels:
I noticed some bugs around `Try` / `TryOption`.

The problem is that some operations (like `Sequence`/`Traverse`) might produce `Try` in `Bottom` state.

`Bottom` state means `Result`/`OptionalResult` will sometimes (`default`) have `Exception == null`.

Many methods like `Match` or `IfFail` don't check this but just pass `Exception` causing possible null errors.

(There was a small bug with order of state evaluation in TryOption, too.)

**Work in progress**: My PR fixes a few things but there is more to do as there are _many_ places to check. Before investing more time I ask for feedback:

a) What is the future of TryOption? Maybe this can be dropped (there is Try<Option>)?
b) What is the future of those Async variants. There are many overloads sometimes (but not complete).

c) Maybe change some things to make implementing the fix easier / code nicer.
One point: never call `Fail(result.Exception)` but create some helper property `Result.ToFail() => Fail(ex ?? new BottomException)` and call `result.ToFail()`.

### DIFF
--- a/LanguageExt.Core/DataTypes/Try/Try.Extensions.cs
+++ b/LanguageExt.Core/DataTypes/Try/Try.Extensions.cs
@@ -113,6 +113,8 @@ public static class TryExtensions
     public static A IfFail<A>(this Try<A> self, Func<Exception, A> Fail)
     {
         var res = self.Try();
+        if (res.IsBottom)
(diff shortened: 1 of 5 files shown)


louthy:
Thanks!

Issue #559, opened 2019-02-27
Comparing Records with polymorphic member fails


slimshader:
When comparing Record<> with polymorphic member (which simulates Discriminated Union), comparison fails with cast exception. Is that the desired behavior? How to handle comparison in such cases? I hoped it would return false in case of different "union case".

Also, I've noticed different behavior when such Records are stored in Map vs direct compare or Lst compare. Maps do not throw, while direct and Lst comparison thows cast exception.

Code to reproduce:
```
using LanguageExt;

namespace ConsoleApp4
{
  interface ArmyState
  {

  }

  class Idle : Record<Idle>, ArmyState
  {
    public Idle(int loc)
    {
      Location = loc;
    }

    public int Location;
  }

  class Dead : Record<Idle>, ArmyState
  {
    public int Location;

    public Dead(int loc)
    {
      Location = loc;
    }
  }

  class Army : Record<Army>
  {
    public int Id { get; }
    public ArmyState State { get; }

    public Army(int id)
    {
      Id = id;
      State = new Idle(5);
    }

    public int Location
    {
      get
      {
        switch (State)
        {
          case Idle idle: return idle.Location;
          case Dead dead: return dead.Location;
          default: return -1;
        }
      }
    }

    private Army(int id, ArmyState state)
    {
      Id = id;
      State = state;
    }

    public Army ToIdle() => new Army(Id, new Idle(Location));
    public Army ToDead() => new Army(Id, new Dead(Location));
  }

  class Program
  {
    static void Main(string[] args)
    {
      var a1 = new Army(1).ToDead();
      var a2 = new Army(1);

      var m1 = new Map<int, Army>().Add(a1.Id, a1);
      var m2 = new Map<int, Army>().Add(a2.Id, a2);


      var mapEq = m1 == m2; // No Exception, true

      var eq = a1 == a2; // No excpetion

      var lstEq = m1.Freeze() == m2.Freeze(); // Exception
    }
  }
}
```

colethecoder:
I've just tried your sample with the latest nuget package and I found this line throws an InvalidCastException:

```cs
var eq = a1 == a2; 
```

but I believe this is because you have:

```cs
class Dead : Record<Idle>, ArmyState
```

which I think should be:

```cs
class Dead : Record<Dead>, ArmyState
```

Adjusting that I then get:

```cs
var mapEq = m1 == m2;               // No Exception, true
var eq = a1 == a2;                  // No Exception, false
var lstEq = 
[...]

PR #561, opened 2019-02-27
Fix bug 559 map equals

bender2k14:
Fixes bug found in #559

`Map.Equals` was returning `true` when it should have returned `false`.  This was because it was checking that the key sets were equal but was ignoring the corresponding values.  I added a line to check for equality of the values using the `object.Equals` instance method on one of the values.  (I checked and `Map` forbids the value to be `null`.)

On the other hand, maybe not checking for equality among values is the intended behavior.

### DIFF
--- a/LanguageExt.Core/DataTypes/Map/Map.Internal.cs
+++ b/LanguageExt.Core/DataTypes/Map/Map.Internal.cs
@@ -1338,6 +1338,7 @@ namespace LanguageExt
                 iterA.MoveNext();
                 iterB.MoveNext();
                 if (!default(OrdK).Equals(iterA.Current.Key, iterB.Current.Key)) return false;
+                if (!iterA.Current.Value.Equals(iterB.Current.Value)) return false;
             }
             return true;
         }
--- a/LanguageExt.Tests/MapTests.cs
+++ b/LanguageExt.Tests/MapTests.cs
@@ -339,6 +339,7 @@ namespace LanguageExtTests
             Assert.False(Map((1, 2)).Equals(emp));
             Assert.False(emp.Equals(Map((1, 2))));
             Assert.True(Map((1, 2)).Equals(Map((1, 2))));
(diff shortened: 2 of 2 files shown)


bender2k14:
This situation is ambiguous because the test were all passing, then I added this line, and all the previous tests continued to pass.  If the behavior in `master` is the intended behavior, then then after rejecting this PR, I would like to create a new PR that adds a test that asserts that two maps with equal key sets but disequal value sets should be considered equal.

slimshader:
This one has bitten me hard too. I switched from Lst<T> to Map<int, T> in my Model implementation and suddenly changes to instances of T (which are Records) are not detected anymore.

louthy:
It was intentional to not include the value in the check, because type `V` might have poor equality support.  This could very well end up being a breaking change for people.  So, let's not do it.

We should probably provide a stronger `Equals` function that takes ` EqV` for passing an equality type-class (`where EqV : struct, Eq<V>`).

bender2k14:
Closing because @louthy said the existing behavior is the intended behavior.  See #570 instead.

StefanBertels:
I know this is an old and closed issue. But IMHO current state isn't very useful in many places and I stumbled upon this, too. 

What are the use cases of having two maps with same keys but different values be equal?

Poor equality of value type is an argument but IMHO we have this problem with all other data types like `Lst`, `Seq`, `Arr`, ....

Maps can be seen as enumerables of key value pairs (in fact they implement `IEnumerable<ValueTuple<K, V>>`) and IMHO  they should have `SequenceEqual` behaviour regarding equality.

`echo-process` uses this code for detecting/publishing actor state changes: 
```
             if (!default(EqDefault<S>).Equals(stateOut, stateIn))
                                    {
                                        stateSubject.OnNext(stateOut);
                                    }
```
If we use e.g. `Hashmap` as state type and just update a value (of a key) this state change will not be detected.

BTW: I don't see a big breaking change. I'm not sure how many would rely on this behavior. I personally would compare `.Keys` property.

Issue #579, opened 2019-04-16
Serialization issue with Newtonsoft


Kavignon:
What to do here? @louthy 
> JsonSerializationException: Cannot create and populate list type LanguageExt.Seq`1

bender2k14:
[`Seq<>`](https://github.com/louthy/language-ext/blob/master/LanguageExt.Core/DataTypes/Seq/Seq.cs#L21) does not have any public constructors, so Newtonsoft is unable to deserialize to that type.

I recommend serializing and deserializing with a type that gives you more control, like [`List<>`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1?view=netframework-4.7.2).

StanJav:
@Kavignon  what lang-ext version are you using? Could you provide some code? We use it a lot for serialisation and it did always work. 

LanguageExt.Core 3.1.18

```
class TestSeq : Record<TestSeq>
{
    public readonly Seq<int> MySeq;

    public TestSeq(Seq<int> mySeq) => MySeq = mySeq;
}

var json = JsonConvert.SerializeObject(new TestSeq(Seq(1, 2, 3)));

// json = {"MySeq":[1,2,3]}
```

bender2k14:
Keep going @StanJav; try to deserialize to `Seq<int>`.

StanJav:
@bender2k14 @Kavignon  Sorry, I was completely wrong. Removed my comment. It is broken.

PR #586, opened 2019-05-02
Made Seq constructor public to support deserialisation and added test

colethecoder:
This allows JSON deserialisation to construct a Seq as highlighted in https://github.com/louthy/language-ext/issues/579


louthy:
Thanks Tom!

Issue #567, opened 2019-03-31
Feature: Resource Monad


gwintering:
This is a proposal to provide more safety around `IDisposable` resources by putting them in  the library's monadic framework.  The inspiration comes from Haskell's [resource monad](https://www.fpcomplete.com/blog/2017/06/understanding-resourcet).

The current `use` family of methods all require you provide a single projection to run with the resource, after which it is disposed.  This pattern has several drawbacks:
1. Despite the lazy, monadic nature of`Try`, you can't return in the `Try` anything that has lazy behavior itself, like a file line iterator or `DbContext` lazy loading proxy.  The problem is that the resource is disposed after producing the first object mapped in the `Try`.
2. The method expects the entire projection at once, which hampers composability.  You can't, for example, call `tryuse` to return a `Try<Connection>` from a method, because the `tryuse` factory also expects you to specify everything you'd like to do with the connection
3. Each method is bespoke logic despite a common pattern
4. We're missing several more use cases (`TryOption`, `TryAsync`, `TryOptionAsync`)

Please take a look at gwintering/language-ext@6ad4ef52 for some sample usage and features.

louthy:
@gwintering I'm not against the idea of a resource monad.  I had problems implementing a general disposal mechanism before with `Try` and other types, so I dropped it.  I am still in the midst of massive time-pressures, so if you want to put something together then I'd be happy to give feedback 👍

louthy:
This is now supported by `Proxy`, which can lift `Aff` and `Eff` monads into its domain.  `Aff` and `Eff` cover the same behaviours (+more) of `Try`, `TryOption`, `TryAsync`, and `TryOptionAsync`;  so I'm going to close this down.

PR #596, opened 2019-06-04
NOT READY Resource Monad

gwintering:
@louthy - I've opened this PR in case you want to provide any feedback on the work so far.  After creating the Resource monad, I realized that it would not participate in the full transformer stack, which made me sad.  So I extended the HKT templates to handle input and output from monads.  Let me know if you want this to be a different PR/issue.

I still need to make a ResourceAsync and update async side of HKT.

closes louthy/language-ext#567

### DIFF
--- a/LanguageExt.Core/DataTypes/Resource/Resource.Prelude.cs
+++ b/LanguageExt.Core/DataTypes/Resource/Resource.Prelude.cs
@@ -1,0 +1,22 @@
+
+using LanguageExt.ClassInstances;
+using LanguageExt.TypeClasses;
+using static LanguageExt.TypeClass;
+using System;
+using System.Diagnostics.Contracts;
+using System.Linq;
+using System.Collections.Generic;
+
+namespace LanguageExt
+{
+    public static partial class Prelude
+    {
+        public static Resource<R> Use<R>(Func<R> acquire) where R : IDisposable =>
+            default(MResource<R>).Use(acquire);
+
+
+        public static Resource<Unit> release<R>(R resource) where R : IDisposable =>
+            default(MResource<R>).Release(
+            default(MResource<R>).Use(() => resource));
+    }
+}
--- a/LanguageExt.Core/HKT/Extensions/TraverseExtensions.cs
+++ b/LanguageExt.Core/HKT/Extensions/TraverseExtensions.cs
@@ -36,54 +36,54 @@ namespace LanguageExt
             return (Value: values, Output: output, IsBottom: false);
         };
 
-        public static Writer<MonoidW, W, Seq<A>> Sequence<MonoidW, W, A>(this Seq<Writer<MonoidW, W, A>> ma) where MonoidW : struct, Monoid<W> =>
-            SequenceFast(ma).Map(Prelude.Seq);
+        // public static Writer<MonoidW, W, Seq<A>> Sequence<MonoidW, W, A>(this Seq<Writer<MonoidW, W, A>> ma) where MonoidW : struct, Monoid<W> =>
+        //     SequenceFast(ma).Map(Prelude.Seq);
 
-        public static Writer<MonoidW, W, Lst<A>> Sequence<MonoidW, W, A>(this Lst<Writer<MonoidW, W, A>> ma) where MonoidW : struct, Monoid<W> =>
-            SequenceFast(ma).Map(toList);
+        // public static Writer<MonoidW, W, Lst<A>> Sequence<MonoidW, W, A>(this Lst<Writer<MonoidW, W, A>> ma) where MonoidW : struct, Monoid<W> =>
+        //     SequenceFast(ma).Map(toList);
 
-        public static Writer<MonoidW, W, Arr<A>> Sequence<MonoidW, W, A>(this Arr<Writer<MonoidW, W, A>> ma) where MonoidW : struct, Monoid<W> =>
-            SequenceFast(ma).Map(toArray);
+        // public static Writer<MonoidW, W, Arr<A>> Sequence<MonoidW, W, A>(this Arr<Writer<MonoidW, W, A>> ma) where MonoidW : struct, Monoid<W> =>
+        //     SequenceFast(ma).Map(toArray);
 
-        public static Writer<MonoidW, W, A[]> Sequence<MonoidW, W, A>(this Writer<MonoidW, W, A>[] ma) where MonoidW : struct, Monoid<W> =>
-            SequenceFast(ma).Map(x => x.ToArray());
+        // public static Writer<MonoidW, W, A[]> Sequence<MonoidW, W, A>(this Writer<MonoidW, W, A>[] ma) where MonoidW : struct, Monoid<W> =>
+        //     SequenceFast(ma).Map(x => x.ToArray());
 
-        public static Writer<MonoidW, W, Set<A>> Sequence<MonoidW, W, A>(this Set<Writer<MonoidW, W, A>> ma) where MonoidW : struct, Monoid<W> =>
-            SequenceFast(ma).Map(toSet);
+        // public static Writer<MonoidW, W, Set<A>> Sequence<MonoidW, W, A>(this Set<Writer<MonoidW, W, A>> ma) where MonoidW : struct, Monoid<W> =>
+        //     SequenceFast(ma).Map(toSet);
 
-        public static Writer<MonoidW, W, HashSet<A>> Sequence<MonoidW, W, A>(this HashSet<Writer<MonoidW, W, A>> ma) where MonoidW : struct, Monoid<W> =>
-            SequenceFast(ma).Map(toHashSet);
+        // public static Writer<MonoidW, W, HashSet<A>> Sequence<MonoidW, W, A>(this HashSet<Writer<MonoidW, W, A>> ma) where MonoidW : struct, Monoid<W> =>
+        //     SequenceFast(ma).Map(toHashSet);
 
-        public static Writer<MonoidW, W, Stck<A>> Sequence<MonoidW, W, A>(this Stck<Writer<MonoidW, W, A>> ma) where MonoidW : struct, Monoid<W> =>
-            SequenceFast(ma).Map(toStack);
+        // public static Writer<MonoidW, W, Stck<A>> Sequence<MonoidW, W, A>(this Stck<Writer<MonoidW, W, A>> ma) where MonoidW : struct, Monoid<W> =>
+        //     SequenceFast(ma).Map(toStack);
 
-        public static Writer<MonoidW, W, IEnumerable<A>> Sequence<MonoidW, W, A>(this IEnumerable<Writer<MonoidW, W, A>> ma) where MonoidW : struct, Monoid<W> =>
-            SequenceFast(ma).Map(Enumerable.AsEnumerable);
+        // public static Writer<MonoidW, W, IEnumerable<A>> Sequence<MonoidW, W, A>(this IEnumerable<Writer<MonoidW, W, A>> ma) where MonoidW : struct, Monoid<W> =>
+        //     SequenceFast(ma).Map(Enumerable.AsEnumerable);
 
 
-        public static Writer<MonoidW, W, Seq<B>> Traverse<MonoidW, W, A, B>(this Seq<Writer<MonoidW, W, A>> ma, Func<A, B> f) where MonoidW : struct, Monoid<W> =>
-            TraverseFast(ma, f).Map(Prelude.Seq);
+        // public static Writer<MonoidW, W, Seq<B>> Traverse<MonoidW, W, A, B>(this Seq<Writer<MonoidW, W, A>> ma, Func<A, B> f) where MonoidW : struct, Monoid<W> =>
+        //     TraverseFast(ma, f).Map(Prelude.Seq);
 
-        public static Writer<MonoidW, W, Lst<B>> Traverse<MonoidW, W, A, B>(this Lst<Writer<MonoidW, W, A>> ma, Func<A, B> f) where MonoidW : struct, Monoid<W> =>
-            TraverseFast(ma, f).Map(toList);
+        // public static Writer<MonoidW, W, Lst<B>> Traverse<MonoidW, W, A, B>(this Lst<Writer<MonoidW, W, A>> ma, Func<A, B> f) where MonoidW : struct, Monoid<W> =>
+        //     TraverseFast(ma, f).Map(toList);
 
-        public static Writer<MonoidW, W, Arr<B>> Traverse<MonoidW, W, A, B>(this Arr<Writer<MonoidW, W, A>> ma, Func<A, B> f) where MonoidW : struct, Monoid<W> =>
-            TraverseFast(ma, f).Map(toArray);
+        // public static Writer<MonoidW, W, Arr<B>> Traverse<MonoidW, W, A, B>(this Arr<Writer<MonoidW, W, A>> ma, Func<A, B> f) where MonoidW : struct, Monoid<W> =>
+        //     TraverseFast(ma, f).Map(toArray);
 
-        public static Writer<MonoidW, W, B[]> Traverse<MonoidW, W, A, B>(this Writer<MonoidW, W, A>[] ma, Func<A, B> f) where MonoidW : struct, Monoid<W> =>
-            TraverseFast(ma, f).Map(x => x.ToArray());
+        // public static Writer<MonoidW, W, B[]> Traverse<MonoidW, W, A, B>(this Writer<MonoidW, W, A>[] ma, Func<A, B> f) where MonoidW : struct, Monoid<W> =>
+        //     TraverseFast(ma, f).Map(x => x.ToArray());
 
-        public static Writer<MonoidW, W, Set<B>> Traverse<MonoidW, W, A, B>(this Set<Writer<MonoidW, W, A>> ma, Func<A, B> f) where MonoidW : struct, Monoid<W> =>
-            TraverseFast(ma, f).Map(toSet);
+        // public static Writer<MonoidW, W, Set<B>> Traverse<MonoidW, W, A, B>(this Set<Writer<MonoidW, W, A>> ma, Func<A, B> f) where MonoidW : struct, Monoid<W> =>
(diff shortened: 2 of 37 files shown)


LanguageExt.Core/DataTypes/Resource/Resource.Prelude.cs
```diff
@@ -0,0 +1,22 @@
...
+using System;
+using System.Diagnostics.Contracts;
+using System.Linq;
+using System.Collections.Generic;
+
+namespace LanguageExt
+{
+    public static partial class Prelude
+    {
+        public static Resource<R> Use<R>(Func<R> acquire) where R : IDisposable =>
```
louthy: Is this supposed to be lower-case? (`use`)

LanguageExt.Core/HKT/Extensions/TraverseExtensions.cs
```diff
@@ -36,54 +36,54 @@ public static class SeqWriterExtensions
             return (Value: values, Output: output, IsBottom: false);
         };
 
-        public static Writer<MonoidW, W, Seq<A>> Sequence<MonoidW, W, A>(this Seq<Writer<MonoidW, W, A>> ma) where MonoidW : struct, Monoid<W> =>
-            SequenceFast(ma).Map(Prelude.Seq);
+        // public static Writer<MonoidW, W, Seq<A>> Sequence<MonoidW, W, A>(this Seq<Writer<MonoidW, W, A>> ma) where MonoidW : struct, Monoid<W> =>
```
louthy: What's the reason for commenting out the `Sequence` extensions?  I don't see an obvious replacement?

LanguageExt.Core/DataTypes/Resource/Resource.Prelude.cs
gwintering: I put it uppercase because it's a type constructor like `Left`, `Fail`, or `Some`.  Maybe `release` below should be uppercase as well?

LanguageExt.Core/HKT/Extensions/TraverseExtensions.cs
gwintering: The method is now generated by the HKT templates.  Github won't let you view it, but the replacement for that method is here https://github.com/gwintering/language-ext/blob/08204cb69dd28a2e5376423d6e0da54854c8c6cd/LanguageExt.Core/HKT/Sync/HKT.Extensions.cs#L69332.  If you're fine with the HKT versions, I'll remove these files in the final PR.

LanguageExt.Core/DataTypes/Resource/Resource.Prelude.cs
louthy: Gah!  Sorry, of course, you're right 👍

LanguageExt.Core/HKT/Extensions/TraverseExtensions.cs
louthy: Fine.  Makes sense.  The only thing that needs to be verified is that the pairing with `IEnumerable` is still lazy.  Also, the `SequenceFast` had (unsurprisngly) fast code, which also stopped unnecessary stack overflows with nested monads; it's quite hard for me to quickly grok the new situation - is there anything you can point me at to make it clearer in my weak mind?

LanguageExt.Core/HKT/OptionT.cs
```diff
@@ -12,73 +12,73 @@ namespace LanguageExt
     /// <typeparam name="OuterType"></typeparam>
```
louthy: This whole file can be deleted.  It was an early experiment, not needed any more.

LanguageExt.Core/TypeClasses/MonadResource/MonadResource.cs
```diff
@@ -0,0 +1,13 @@
+using LanguageExt.Attributes;
+using System;
+
+namespace LanguageExt.TypeClasses
+{
+    [Typeclass]
+    public interface MonadResource : Typeclass
```
louthy: I notice `Use` and `Release` both have a generic of `R` - shouldn't the interface be `MonadResoure<R>` - is there a benefit to applying it on the methods?

louthy: I think I see now.  `MonadResource<R>` would force anything implementing it to have its generic parameter constrained to `IDisposable`, and so this allows _bolting on_ of `Use` and `Release` to any monadic type.

louthy:
@gwintering So the `MonadResource` definition was still bugging me a bit.  I'm not sure if I've missed the point a bit here, but I tried reformulating it to be as general as possible:
```c#
    [Typeclass]
    public interface MonadResource<MIn, MOut, MonadA, MA, A, MonadUnit, MU> : Typeclass
        where MonadA : struct, Monad<MIn, MOut, MA, A>
        where MonadUnit : struct, Monad<MIn, MOut, MU, Unit>
    {
        MA Use(Func<A> acquire);
        MU Release(MA resource);
    }

    [Typeclass]
    public interface MonadResource<MonadA, MA, A, MonadUnit, MU> : 
        MonadResource<Unit, Unit, MonadA, MA, A, MonadUnit, MU>
        where MonadA : struct, Monad<Unit, Unit, MA, A>
        where MonadUnit : struct, Monad<Unit, Unit, MU, Unit>
    {
    }
```
NOTE: There's no `IDisposable` constraint on `A`.

Then the declaration of `MResource<A>` looks like this:
```c#
public struct MResource<A> :
    MonadResource<Unit, HashMap<IDisposable, bool>, MResource<A>, Resource<A>, A, MResource<Unit>, Resource<Unit>>,
    Monad<Unit, HashMap<IDisposable, bool>, Resource<A>, A> 
{
    ...

    public Resource<Unit> Release(Resource<A> ma) => () =>
    {
        var (released, acquired, faulted) = ma();
        return released is IDisposable disp
            ? (unit, acquired.SetItem(disp, false), faulted)
            : (unit, acquired, faulted);
    };

    public Resource<A> Use(Func<A> acquire) =>
        () =>
        {
            var acquired = acquire();
            return (acquired, acquired is IDisposable disp ? HashMap((disp, true)) : HashMap<IDisposable, bool>(), false);
        };

    ...
}

```

We could also let `Release` return the original `A` ... it's disposed, but still a valid object - not ideal, but it would allow the above to be simplified to:
```c#
    [Typeclass]
    public interface MonadResource<MIn, MOut, MonadA, MA, A> : Typeclass
        where MonadA : struct, Monad<MIn, MOut, MA, A>
    {
        MA Use(Func<A> acquire);
        MA Release(MA resource);
    }

    [Typeclass]
    public interface MonadResource<MonadA, MA, A> : 
        MonadResource<Unit, Unit, MonadA, MA, A>
        where MonadA : struct, Monad<Unit, Unit, MA, A>
    {
    }
```
With `MResource` looking like this:
```c#
    public struct MResource<A> :
        MonadResource<Unit, HashMap<IDisposable, bool>, MResource<A>, Resource<A>, A>,
        Monad<Unit, HashMap<IDisposable, bool>, Resource<A>, A> 
    {

        public Resource<A> Run(Func<Unit, Resource<A>> ma) => () =>
            ma(unit)();

        public Resource<A> Release(Resource<A> ma) => () =>
        {
            var (released, acquired, faulted) = ma();
            return released is IDisposable disp
                ? (released, acquired.SetItem(disp, false), faulted)
                : (released, acquired, faulted);
        };
   }
```
Is this of value?  Or, is it just adding complexity where it's not needed?  There's still a bit of brain fog that I need to clear.

LanguageExt.Core/HKT/Extensions/TraverseExtensions.cs
gwintering: Do you have the same concerns about, for example, `Seq<Option<>>.Sequence()` or `Seq<Try<>>.Sequence()`?  I ask because this change just allows types with `Env` and `Out` to participate in the same transformers as the rest of the monads.  So if `SequenceFast` is needed to ensure laziness or stack elimination, it means most of the `MA<MB<>>.Sequence()` calls suffer from those problems because they **do not** use `SequenceFast`.

Which part of `SequenceFast` is the critical part, the `foreach`?  If so we're good because the new version uses `SeqTrans<>.Traverse`, which ends up calling `Seq<>.Fold`, which performs the accumulation inside a `foreach`.

If the critical part of `SequenceFast` is accumulating the values on a `List<>`, then this change may have an impact because we're now creating the accumulation with a combination of `MSeq<>.Plus` and `MSeq<>.Return`.  But 2bf43ab makes `Seq<>` efficient anyway so I doubt we see an impact anyway.

louthy: > So if SequenceFast is needed to ensure laziness or stack elimination

I'm trying to rack my brains for the issues that were raised on this.  There were a couple, one was performance related (with nested monads) and one was stack-overflows with nested monads (one being an `IEnumerable` I believe).  I think the stack overflows were solved with `EnumerableOptimal.ConcatFast` which is used in `Append` and `Plus` - that may have been the fix for perfomance too, but I can't be certain right now (the commit you linked to is just version change?).

> Do you have the same concerns about, for example, Seq<Option<>>.Sequence() or Seq<Try<>>.Sequence()?

This may all be a red herring and I'm just misremembering - I've got about 5 different projects in my head right now, and context switching is difficult;  I obviously want to make sure we don't have some massive drop in performance or a return to previous bugs [because of the change].  

The laziness aspect should just be confirmed with a unit test though.  I've had a couple of bugs over the years, with the transformer system, where I'd accidentally forced a lazy collection to enumerate.

I am doing some major performance improvements on another branch anyway - so this might all just be moot!

LanguageExt.Core/TypeClasses/MonadResource/MonadResource.cs
gwintering: I actually didn't really consider applying `MonadResource` to other types, but I like the idea and I think that's what you're working on below.

gwintering:
I like generalizing `MonadResource<>` to work like `Choice<>`; I didn't consider that and was just cargo culting `MonadWriter`.

It's good that we don't have a global condition `where A : IDisposable` because the idea is that you work within the monad (lazily) till you map to a `Resource<B>` where `B` is not `IDisposable`, then you `.Run()` it to actually produce a `B` (and `Dispose` all resources).  If `B` is `IDisposable`, `.Run()` will dispose of it so it won't be of much use.

For example, we might acquire a `DbContext` and then a `FileStream` and use data within them to finally yield a `decimal` price, at which point we'd have a `Resource<decimal>` on which we would call `Run` to get a `decimal`.

Although `MonadUnit` and `MU` are ugly, it's worse for `Release` to return `MA`.  Only the designer of the class instance has to worry about those type arguments so the ugliness is contained.  In return for that ugliness, we get an honest and poka yoke API that doesn't return an object that will blow up if used.

louthy:
> Although MonadUnit and MU are ugly, it's worse for Release to return MA

I mostly agree, however...

>  Only the designer of the class instance has to worry about those type arguments so the ugliness is contained.

I think this isn't true if we generalise it.  I'm quite mindful of the functions that would need to not only specify `M<A>` but also `M<Unit>` in their constraints - this additional `M<Unit>` is less likely to have been passed through a nested chain of generic calls - and therefore makes `MonadResource` harder to use (it feels annoyingly arbitrary):
```c#
public static Unit Foo<MonadA, MA, A>(MA ma)
    where MonadA : MonadResource<MonadA, MA, A, /* From where MonadUnit? */, /* From where MU? */>,
                   Monad<Unit, HashMap<IDisposable, bool>, MA, A> =>
{

}
```
I realise that the _correct_ argument is that we should be type-safe all the way through, but I think there's an argument of pragmatism here.  The potential pragmatic decision here is to go with the `A` return (as mentioned before).
```c#
    public static Foo<MonadA, MA, A>(MA ma)
        where MonadA : MonadResource<MonadA, MA, A>,
                       Monad<Unit, HashMap<IDisposable, bool>, MA, A> =>
    {

    }
```
`Release` is not returning `null` although I accept that a disposed object is almost as bad.  I'm really in two minds here.  I'd prefer to return `M<Unit>` obviously, it just feels very heavyweight in real use.

And also we should probably create a box for `HashMap<IDisposable, bool>`.  One that's likely to be more performant.  For now it can wrap `HashSet<A>` (I wasn't quite sure why it had to be a `HashMap` with `true` and `false` - it seems `true` is _in-the-set_ and `false` is _not-in-the-set_?  But, perhaps I'm missing something:
```c#
    public struct Resources : IDisposable
    {
        public readonly static Resources Empty = new Resources(HashSet<IDisposable>());

        readonly HashSet<IDisposable> resources;

        Resources(HashSet<IDisposable> resources) =>
            this.resources = resources;

        public static Resources Acquire(IDisposable disp) =>
            new Resources(HashSet(disp));

        public Resources Release(IDisposable disp) =>
            new Resources(resources.Remove(disp));

        public Resources Merge(Resources rb) =>
            new Resources(resources.Union(rb.resources));

        public void Dispose() =>
            resources.Iter(r => r?.Dispose());

        public int Count => 
            resources.Count;
    }
```
I suspect `Resources` would be better with an array internally, because of the smaller number of resources likely to be held per operation.  But, at least by wrapping it, we can optimise later without infecting other code.

Then `Foo` starts to look a bit more elegant:
```c#
    public static Foo<MonadA, MA, A>(MA ma)
        where MonadA : MonadResource<MonadA, MA, A>,
                       Monad<Unit, Resources, MA, A> =>
    {

    }
```

gwintering:
> The more specific implementations like Resource<R> can map away the A

Good point there.  Whatever we decide we still have that final level of indirection that can avoid giving users a landmine.

Just to make sure I'm understanding you correctly, the `Foo` function is something that would be in a `MonadResource.Prelude.cs`?  I hadn't considered `Prelude` yet but that is a gap I'll fill.

If we keep the unit return, the way to make `Foo` work would be for it to demand monad unit type arguments from its callers
```csharp
// This is in MonadResource.Prelude
public static Unit Foo<MonadA, MA, A, MonadU, MU>(MA ma)
    where MonadA : MonadResource<MonadA, MA, A, MonadU, MU>,
                   Monad<Unit, HashMap<IDisposable, bool>, MA, A> =>
```
You are correct that _within_ `MonadResource.Prelude.cs` there would be some additional `MonadU`s and `MU`s when the functions there called each other.

For the individual data type(i.e. `Resource<A>`) it doesn't look that bad:
```csharp
// This in in Resource.Extensions
public static Unit Foo(this Resource<A>, MA ma) =>
  MonadResource.Foo<MResource<MResource<A>, Resource<A>, A, MResource<Unit>, Resource<Unit>>(ma);
```

Providing a `MonadU`(e.g. `MResource<Unit>`) and `MU`(e.g. `Resource<Unit>`) is as easy as providing a `MonadA`(e.g. `MResource<A>`) and `MA`(e.g. `Resource<A>`).

Just for reference, the situation I was considering was adding the `MonadResource` typeclass to a class instance:
```csharp
    public struct MTryResource<A> :
        MonadResource<Unit, HashMap<IDisposable, bool>, MTryResource<A>, TryResource<A>, A, MTryResource<Unit>, TryResource<Unit>>,
        Choice<TryResource<A>, Exception, A>,
        Monad<Unit, HashMap<IDisposable, bool>, TryResource<A>, A>
    {

```
Here the `MonadU` and `MU` are not difficult to provide.

You've got more experience designing this library for super generic purposes, so if your intuition says `MonadU` and `MU` are going to get in the way we can leave them out.  On the other hand there is something fundamental about `SomeMonad<Unit>` so maybe `MonadU` and `MU` are an avenue for even more generality in the future.

I'm also fine with doing the whole feature with `MonadU` and `MU` and having you review it then and decide whether to keep them.  At least with that direction we would have concrete code to talk about.

> And also we should probably create a box for HashMap<IDisposable, bool>

Agreed

> I wasn't quite sure why it had to be a HashMap with true and false - it seems true is in-the-set and false is not-in-the-set?

Close.  The boolean value was whether the resource was available.  I was considering having `Release` complain if you asked it to release an unknown resource but tolerate rereleasing a known resource, i.e. release of known resources is idempotent.  It's also a bit easier to debug with because you get the complete picture of all the resources it's controlling.  But I understand why you'd want `HashSet` there and we can decide that during perf profiling.

louthy:
> Just to make sure I'm understanding you correctly, the Foo function is something that would be in a MonadResource.Prelude.cs? I hadn't considered Prelude yet but that is a gap I'll fill.

No, `Foo` is an example of a function written by someone using this library.  The constraints are onerous if they have to specify `M<Unit>` and `M<A>`.  I realise that probably very few people will actually be using this ad-hoc polymorphic framework (because the generics story is so ugly), but it would be nice to minimise the pain somewhat, especially when we can bury that pain for types like `Resource<A>`.  So, mostly I'm trying to think of possible end-users here.

> If we keep the unit return, the way to make Foo work would be for it to demand monad unit type arguments from its callers

I understand that, but for the generic ad-hoc polymorphic story to work at all, there has to be a _beginning_ which is typed, and as those types are passed in they have to be threaded through every nested level of functionality.  It becomes more difficult and more onerous if one has to specify `M<A>` and `M<Unit>`.  It's just an uglier experience for the user basically.

>  On the other hand there is something fundamental about SomeMonad<Unit> so maybe MonadU and MU are an avenue for even more generality in the future.

I was considering this too, and spent an hour or so thinking about it before being dragged into other work.  I was thinking instead of returning `M<Unit>` we could return `MonadUnit` - which represents a general unit value that any monad is implicitly convertible from.  i.e.

```c#
    M<A> -> converts to -> MonadUnit -> converts to -> M<Unit>
```
I then realised that `MonadUnit` is essentially just `Unit` and it couldn't ever really auto-convert to say `Seq<A>`, because `Unit -> Seq<Unit>` would have to create an empty `Seq<Unit>` which represents a *failed* state.  

I think on reflection let's go with having to specify `M<Unit>` - it's easier to remove that constraint later if it turns out to be too painful than to re-add it later.  I think the chances of anyone actually using this stuff in their code is minimal and therefore this has more value as a tool for constructing lang-ext types - which should demand correctness where reasonable and appropriate.

louthy:
https://github.com/louthy/language-ext/issues/567#issuecomment-929325286

PR #607, opened 2019-07-10
 Add Angstrom to Length UnitOfMeasure

rantri:
"Angstrom" is a very popular length measure used in semiconductor industry.


louthy:
Thanks @rantri :+1:

PR #608, opened 2019-07-15
multithreading issue with static parser definition (eof)

StefanBertels:
This test fails with a NullReferenceException. I guess this is caused by static eof using static anychar being null in some execution path.

### DIFF
--- a/LanguageExt.Tests/ParsecTests.cs
+++ b/LanguageExt.Tests/ParsecTests.cs
@@ -552,6 +552,10 @@ namespace LanguageExtTests
         [Fact]
         public void ParallelCheck()
         {
+            // works
+            Parallel.ForEach(Enumerable.Repeat("", 4), str => parse(from _ in notFollowedBy(anyChar).label("end of input") select unit, str));
+            
+            // sometimes crashes (net461)
             Parallel.ForEach(Enumerable.Repeat("", 4), str => parse(from _ in eof select unit, str));
         }
     }
--- a/LanguageExt.Tests/ParsecTests.cs
+++ b/LanguageExt.Tests/ParsecTests.cs
@@ -548,5 +548,11 @@ namespace LanguageExtTests
             Assert.True(r4.IfLeft("x") == "1234");
             Assert.True(r5.IfLeft("x") == "1234");
         }
+        
+        [Fact]
+        public void ParallelCheck()
+        {
+            Parallel.ForEach(Enumerable.Repeat("", 4), str => parse(from _ in eof select unit, str));
+        }
     }
 }


StefanBertels:
Note: this just shows the bug, no bugfix included. (I'm not sure how to fix this properly.)

Test does not fail always (hard to reproduce). Seems like running `dotnet test` is more useful than unit test runners in development.

bender2k14:
I couldn't get your commit to compile.  There was a syntax error in the auto-generated code.  Your commit compiles for me after I cherry pick it at tag `v3.1.25`.

bender2k14:
As you suggested, you test is always passing for me when running from within Visual Studio.  Can you post a screenshot of it failing?

StefanBertels:
Compilation of the whole project raises errors here, too. 

But the test projects does not need those parts (here).

The test seems to fail on net461 only: 

`dotnet test --filter "LanguageExtTests.ParsecTests.ParallelCheck" -f net461`

Here's a my log (for both target frameworks):

```
Test run for D:\github\louthy\language-ext\LanguageExt.Tests\bin\Debug\netcoreapp2.0\LanguageExt.Tests.dll(.NETCoreApp,Version=v2.0)
Microsoft (R) Test Execution Command Line Tool Version 16.0.1
Copyright (c) Microsoft Corporation.  All rights reserved.

Starting test execution, please wait...

Total tests: 1. Passed: 1. Failed: 0. Skipped: 0.
Test Run Successful.
Test execution time: 3,2804 Seconds
Test run for D:\github\louthy\language-ext\LanguageExt.Tests\bin\Debug\net461\LanguageExt.Tests.dll(.NETFramework,Version=v4.6.1)
Microsoft (R) Test Execution Command Line Tool Version 16.0.1
Copyright (c) Microsoft Corporation.  All rights reserved.

Starting test execution, please wait...
[xUnit.net 00:00:02.18]     LanguageExtTests.ParsecTests.ParallelCheck [FAIL]
Failed   LanguageExtTests.ParsecTests.ParallelCheck
Error Message:
 System.AggregateException : One or more errors occurred.
---- System.NullReferenceException : Object reference not set to an instance of an object.
---- System.NullReferenceException : Object reference not set to an instance of an object.
Stack Trace:
   at System.Threading.Tasks.Task.ThrowIfExceptional(Boolean includeTaskCanceledExceptions)
   at System.Threading.Tasks.Task.Wait(Int32 millisecondsTimeout, CancellationToken cancellationToken)
   at System.Threading.Tasks.Task.Wait()
   at System.Threading.Tasks.Parallel.PartitionerForEachWorker[TSource,TLocal](Partitioner`1 source, ParallelOptions parallelOptions, Action`1 simpleBody, Action`2 bodyWithState, Action`3 bodyWithStateAndIndex, Func`4 bodyWithStateAndLocal, Func`5 bodyWithEverything, Func`1 localInit, Action`1 localFinally)
   at System.Threading.Tasks.Parallel.ForEachWorker[TSource,TLocal](IEnumerable`1 source, ParallelOptions parallelOptions, Action`1 body, Action`2 bodyWithState, Action`3 bodyWithStateAndIndex, Func`4 bodyWithStateAndLocal, Func`5 bodyWithEverything, Func`1 localInit, Action`1 localFinally)
   at System.Threading.Tasks.Parallel.ForEach[TSource](IEnumerable`1 source, Action`1 body)
   at LanguageExtTests.ParsecTests.ParallelCheck() in D:\github\louthy\language-ext\LanguageExt.Tests\ParsecTests.cs:line 556
----- Inner Stack Trace #1 (System.NullReferenceException) -----
   at LanguageExt.Parsec.Prim.<>c__DisplayClass18_0`1.<attempt>b__0(PString inp) in D:\github\louthy\language-ext\LanguageExt.Parsec\Parsers\Prim.cs:line 246
   at ParserExtensions.<>c__DisplayClass8_0`3.<SelectMany>b__0(PString inp) in D:\github\louthy\language-ext\LanguageExt.Parsec\Parser.cs:line 108
   at LanguageExt.Parsec.Prim.<>c__DisplayClass13_0`1.<either>b__0(PString inp) in D:\github\louthy\language-ext\LanguageExt.Parsec\Parsers\Prim.cs:line 134
   at LanguageExt.Parsec.Prim.<>c__DisplayClass18_0`1.<attempt>b__0(PString inp) in D:\github\louthy\language-ext\LanguageExt.Parsec\Parsers\Prim.cs:line 246
   at ParserExtensions.<>c__DisplayClass1_0`1.<label>b__0(PString inp) in D:\github\louthy\language-ext\LanguageExt.Parsec\Parser.cs:line 58
   at ParserExtensions.<>c__DisplayClass7_0`2.<Select>b__0(PString inp) in D:\github\louthy\language-ext\LanguageExt.Parsec\Parser.cs:line 98
   at ParserExtensions.Parse[T](Parser`1 self, String input) in D:\github\louthy\language-ext\LanguageExt.Parsec\Parser.cs:line 81
   at LanguageExt.Parsec.Prim.parse[T](Parser`1 p, String input) in D:\github\louthy\language-ext\LanguageExt.Parsec\Parsers\Prim.cs:line 28
   at LanguageExtTests.ParsecTests.<>c.<ParallelCheck>b__44_1(String str) in D:\github\louthy\language-ext\LanguageExt.Tests\ParsecTests.cs:line 556
   at System.Threading.Tasks.Parallel.<>c__DisplayClass42_0`2.<PartitionerForEachWorker>b__1()
   at System.Threading.Tasks.Task.InnerInvoke()
   at System.Threading.Tasks.Task.InnerInvokeWithArg(Task childTask)
   at System.Threading.Tasks.Task.<>c__DisplayClass176_0.<ExecuteSelfReplicating>b__0(Object )
----- Inner Stack Trace #2 (System.NullReferenceException) -----
   at LanguageExt.Parsec.Prim.<>c__DisplayClass18_0`1.<attempt>b__0(PString inp) in D:\github\louthy\language-ext\LanguageExt.Parsec\Parsers\Prim.cs:line 246
   at ParserExtensions.<>c__DisplayClass8_0`3.<SelectMany>b__0(PString inp) in D:\github\louthy\language-ext\LanguageExt.Parsec\Parser.cs:line 108
   at LanguageExt.Parsec.Prim.<>c__DisplayClass13_0`1.<either>b__0(PString inp) in D:\github\louthy\language-ext\LanguageExt.Parsec\Parsers\Prim.cs:line 134
   at LanguageExt.Parsec.Prim.<>c__DisplayClass18_0`1.<attempt>b__0(PString inp) in D:\github\louthy\language-ext\LanguageExt.Parsec\Parsers\Prim.cs:line 246
   at ParserExtensions.<>c__DisplayClass1_0`1.<label>b__0(PString inp) in D:\github\louthy\language-ext\LanguageExt.Parsec\Parser.cs:line 58
   at ParserExtensions.<>c__DisplayClass7_0`2.<Select>b__0(PString inp) in D:\github\louthy\language-ext\LanguageExt.Parsec\Parser.cs:line 98
   at ParserExtensions.Parse[T](Parser`1 self, String input) in D:\github\louthy\language-ext\LanguageExt.Parsec\Parser.cs:line 81
   at LanguageExt.Parsec.Prim.parse[T](Parser`1 p, String input) in D:\github\louthy\language-ext\LanguageExt.Parsec\Parsers\Prim.cs:line 28
   at LanguageExtTests.ParsecTests.<>c.<ParallelCheck>b__44_1(String str) in D:\github\louthy\language-ext\LanguageExt.Tests\ParsecTests.cs:line 556
   at System.Threading.Tasks.Parallel.<>c__DisplayClass42_0`2.<PartitionerForEachWorker>b__1()
   at System.Threading.Tasks.Task.InnerInvoke()
   at System.Threading.Tasks.Task.InnerInvokeWithArg(Task childTask)
   at System.Threading.Tasks.Task.<>c__DisplayClass176_0.<ExecuteSelfReplicating>b__0(Object )

Total tests: 1. Passed: 0. Failed: 1. Skipped: 0.
Test Run Failed.
Test execution time: 2,8699 Seconds
```

StefanBertels:
Some more notes: See advanced test for comparison. Substantiates my thesis initialization of static fields eof/anyChar might be a problem (with .NET framework). 

I detected the issue in a project using LanguageExt.Parsec as package reference. I could not reproduce this in another project (similar setup, i.e. LanguageExt as packag reference) -- the test above runs fine there (even in net461) -- with 3.1.24 and 3.2.6-beta. So it's a bit hard to reproduce. 

But I hope the original test in LanguageExtTests.ParsecTests (commit above) will be sufficient to show it.

louthy:
@StefanBertels I have pushed a fix in `3.2.7-beta`

StefanBertels:
Thanks for fixing it really fast! Does anybody know the exact reason for the broken behaviour?

Is it a bug in .NET runtime or just undefined? The fix looks good to me -- at least I found this regarding constructors: https://github.com/dotnet/docs/issues/10243#issuecomment-467222385

louthy:
The order of C# static member field initialisation is 'undefined' in the C# spec.  But in my experience it has always gone from the member at the top of the class, in order, down to the bottom (and so items at the bottom can refer to items above, but not vice versa).  

In the case of a dependency graph of a static field in one class referring to a static field in another, then perhaps there's a race condition in the static initialisation of members, which your test highlighted.  I figured that, because it seemed hard to reproduce, then the pertinent thing to do would be:

1. Enforce the static initialisation order within the class by moving the initialisation into the constructor
2. Remove the cross class link for `eof` seeing as it was a very simple parser

I didn't prove it, but perhaps the race condition would go away because of 1.  Nothing should be able to access a member until the constructor has completed, and therefore there can't be a race condition 

> Obviously this is making an assumption that there is in fact a race condition, and not some other cause).

StefanBertels:
Just a comment because I remembered this issue when reading about static constructors in the docs. I found this showing that initialisation order seems to be defined:

> A static constructor is called automatically to initialize the class before the first instance is created or any static members are referenced. A static constructor will run before an instance constructor. A type's static constructor is called when a static method assigned to an event or a delegate is invoked and not when it is assigned. **If static field variable initializers are present in the class of the static constructor, they will be executed in the textual order in which they appear in the class declaration** immediately prior to the execution of the static constructor.

Source: https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/static-constructors

and

> The default value initialization described in Field initialization occurs for all fields, including fields that have variable initializers. Thus, when a class is initialized, all static fields in that class are first initialized to their default values, and **then the static field initializers are executed in textual order**. Likewise, when an instance of a class is created, all instance fields in that instance are first initialized to their default values, and then the instance field initializers are executed in textual order.

Source: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/classes#variable-initializers

I guess the cross class link issue was relevant for the bug. `Char` calls `Prim`, the fix removed the other direction (`Prim.eof` called `Char`). It's still weird that this causes issues only in multi threaded code. Perhaps compiler avoids dead locks by allowing calls to uninitialized fields. Seems it's best to avoid accessing other classes at all (or have a clear hierarchy).

Having explicit initialization order by static constructor is a good thing anyway -- getting safety when fields refer to each other, especially if a developer (or tool) accidentally reorders field.

PR #609, opened 2019-07-17
Fix inconsistency in LanguageExt.Tests namespace

rantri:
Change namespace from LanguageExtTests to LanguageExt.Tests in some of the files


louthy:
Thanks @rantri !

PR #621, opened 2019-08-14
HashMap implementation is broken after RemoveRange()

iamim:
I noticed that `HashMap` is broken with certain hash values of `Key` after `RemoveRange()` operation.

### DIFF
--- a/LanguageExt.Tests/HashMapTests.cs
+++ b/LanguageExt.Tests/HashMapTests.cs
@@ -227,5 +227,22 @@ namespace LanguageExt.Tests
             Assert.True(HashMap<int, int>((3, 4), (1, 2)).Equals(HashMap<int, int>((1, 2), (3, 4))));
             Assert.True(HashMap<int, int>((3, 4), (1, 2)).Equals(HashMap<int, int>((3, 4), (1, 2))));
         }
+        
+        [Fact]
+        public void FetchBack()
+        {
+            var init = Seq(69, 1477);
+            var rmv = Seq1(69);
+
+            var map = toHashMap(init.Zip(Enumerable.Repeat(1, int.MaxValue)));
+            var minus = map.RemoveRange(rmv);
+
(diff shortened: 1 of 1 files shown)


bender2k14:
I confirm this.

Stepping through the debugger, it seems to me that the issue is in this [`TrieMap.Read`](https://github.com/louthy/language-ext/blob/master/LanguageExt.Core/DataTypes/TrieMap/TrieMap.cs#L1096) method.  In particular, the instance state seems correct to me but instead this method [returns `default` here](https://github.com/louthy/language-ext/blob/master/LanguageExt.Core/DataTypes/TrieMap/TrieMap.cs#L1126).

iamim:
@bender2k14 I think you may be right. Stepping through it yesterday I noticed that `EntryMap` value of `TrieMap` was weird.
I discovered it by testing my solution with FsCheck. I ran some more testing this morning by explicitly copying all `HashMap` after `RemoveRange` but the problem is still there. It seems like there are more cases where the implementation breaks. I can't say yet what's the cause.
If @louthy thinks there's a value in it we could create a FsCheck test project to test cases like this. This library does magic.
As a side note, I created this PR late at night. It should've been an issue instead. Don't know how to change it.

bender2k14:
Is it necessary to create a new test project in order to write property-based testing with FsCheck?

iamim:
@bender2k14 no, it works perfectly fine with an existing NUnit/xUnit project. My thinking was that this kind of tests may run for a much longer time. Sometimes I create a FsCheck test and click "Run until fails" in my IDE.
It may suffice to categorize property tests as long-running so that they are not run together with other tests. Or it may be just fine as is, it's up to maintainers to decide.

bender2k14:
That makes perfect sense.  I didn't know "Run until fails" was an option.  Neat.

louthy:
@iamim All additional unit tests are welcome.  Anything that gives more confidence is good.

Issue #622, opened 2019-08-16
CodeGen fails to generate With for Records with field backed properties


koffmoff:
To my understanding, both of the below `Record` implementations should be valid.

```cs
    [With]
    public partial class Works : Record<Works>
    {
        public readonly String A;
        
        public Works(string a)
        {
            this.A = a;
        }
    }

    [With]
    public partial class Fails : Record<Fails>
    {
        public string A { get; }

        public Fails(string a)
        {
            this.A = a;
        }
    }
```

However, the CodeGen-generated code differs:

```cs
    public partial class Works
    {
        public Works With(String A = null) => new Works(A ?? this.A);
    }

    public partial class Fails
    {
        public Fails With() => new Fails();
    }
```

Which obviously doesn't compile since `Fails` does not have a default constructor.

A CodeGen bug or some misunderstanding on my behalf?

louthy:
Yes, it's very much set up to work with `readonly` fields rather than field-backed properties which have a nasty habit of not necessarily being setup in a constructor.

So, use `public readonly string A` instead of `public string A { get; }`.  

There's probably more that could be done to spot the use of auto-field backed properties, but there's also a value to keeping the analyser fast and simple.

koffmoff:
The reason I wan't to use properties is I have a WPF application and WPF can only bind to properties. Now I have to wrap my Records in other, WPF-friendly classes which is not very nice.

I though it'd be a simple fix but I suspect from your answer that you've already been down this road and that it's more complicated than it seems. Care to give some pointers and maybe I could take a look if I could find some time?

louthy:
>  Care to give some pointers and maybe I could take a look if I could find some time?

No problem, it's not super hard to do this, it's just a bit of experimentation and typing (i.e. time, which isn't always available).

[This function](https://github.com/louthy/language-ext/blob/master/LanguageExt.CodeGen/CodeGenUtil.cs#L14) returns a `List<FieldDeclarationSyntax> Fields`, which is built from:
```c#
var fields = applyToClass.Members
                         .Where(m => m is FieldDeclarationSyntax)
                         .Select(m => m as FieldDeclarationSyntax)
                         .Where(f => f.Declaration.Variables.Count > 0)
                         .Where(f => FirstCharIsUpper(f.Declaration.Variables[0].Identifier.ToString()))
                         .Where(f => f.Modifiers.Any(SyntaxKind.PublicKeyword))
[...]

PR #625, opened 2019-08-18
CodeGen of With/Lens should work for properties

koffmoff:
As per #622

### DIFF
--- a/LanguageExt.CodeGen/CodeGenUtil.cs
+++ b/LanguageExt.CodeGen/CodeGenUtil.cs
@@ -11,7 +11,7 @@ namespace LanguageExt.CodeGen
 {
     internal static class CodeGenUtil
     {
-        public static (ClassDeclarationSyntax PartialClass, TypeSyntax ReturnType, List<FieldDeclarationSyntax> Fields) GetState(TransformationContext context)
+        public static (ClassDeclarationSyntax PartialClass, TypeSyntax ReturnType, List<(SyntaxToken Identifier, TypeSyntax Type, SyntaxTokenList Modifiers)> Fields) GetState(TransformationContext context)
         {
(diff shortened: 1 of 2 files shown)


LanguageExt.CodeGen/CodeGenUtil.cs
```diff
@@ -38,16 +38,38 @@ public static (ClassDeclarationSyntax PartialClass, TypeSyntax ReturnType, List<
...
+                                         .Where(p => p.AccessorList.Accessors[0].ExpressionBody == null)
+                                         .Where(p => p.AccessorList.Accessors[0].Body == null)
+                                         .Where(p => p.Initializer == null)
+                                         .Select(p => (
+                                             p.Identifier,
+                                             p.Type,
+                                             p.Modifiers
+                                         ));
+ 
+            var fields = variables.Concat(properties).ToList();
```
louthy: This loses the ordering of fields and properties inthe class.  The guidance is that the order the fields are in the class are the order of the entries to the constructor.  Anyone mixing the order of the fields and properties would be confused at this.

One technique might be to zip the members with an `Enumerable.Range(0, Int32.MaxValue)` to create an index for each member and then use that index to sort the final set of results. 

It'd be good to rename `variables`  to fields and `fields` to `members` also.

koffmoff: Done. A bit redundant to impose restrictions on both naming and ordering of the constructor argument, don't you think? But that's out of topic here.

louthy:
Looks good, thanks Fredrik.  I'll try and deploy to nu-get tomorrow. 👍

koffmoff:
Great. Keep up the good work. Much appreciated!

PR #629, opened 2019-08-30
Fixed - Bug in ReaderResult Match

nicholasham:
**Test that should surface the bug**

```c#
using Xunit;

namespace LanguageExt.Tests
{
    public class ReaderResultTests
    {
        [Fact]
        public void MatchWhenNotFaulted()
        {
            // This will throw an InvalidCastException
            var expected = "Some value";
            var actual = ReaderResult<string>.New(expected).Match(s => s, error => error.Message);
            Assert.Equal(expected, actual);
        }
        
        [Fact]
        public void MatchWhenFaulted()
        {
            var expected = "Some error";
            var actual = ReaderResult<string>.New(Common.Error.New(expected)).Match(s => s, error => error.Message);
            Assert.Equal(expected, actual);
        }
        
        
    }
}
```

### DIFF
--- a/LanguageExt.Core/DataTypes/Reader/Reader.cs
+++ b/LanguageExt.Core/DataTypes/Reader/Reader.cs
@@ -46,8 +46,8 @@ namespace LanguageExt
 
         public B Match<B>(Func<A, B> Succ, Func<Error, B> Fail) =>
             IsFaulted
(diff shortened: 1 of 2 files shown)


louthy:
@nicholasham Thanks for this, sorry I missed it and had done the fix myself.

Issue #635, opened 2019-09-19
Seq.Add broken - returns the same result for different inputs


StanJav:
```
var s = Seq1(0);
var x = s.Add(1);
var z = s.Add(2);

Console.WriteLine(x.Last());
Console.WriteLine(z.Last());
```

Expected: 12
Current: 11 <-- so second `Add` did something wrong

I have tested versions 3.3.16, 3.3.24 and 3.3.25-beta

colethecoder:
The problem is that CloneAdd (https://github.com/louthy/language-ext/blob/9e18319195ae286bd99124fd4c82849dc4dad993/LanguageExt.Core/DataTypes/Seq/SeqStrict.cs#L180) for SeqStrict (which is being used here behind the scenes) always assumes it needs to grow the array before cloning and that the next entry will go at the end of the current array.

I think it's a pretty easy fix I'll make a PR later.

StanJav:
Thanks for looking at it @colethecoder

louthy:
Fixed in `3.3.26` thanks @colethecoder

PR #636, opened 2019-09-19
Test and fix for Seq.Add issue

colethecoder:
In relation to #635 

Seq.Add calls CloneAdd when the Seq has already been added to. CloneAdd assumed the array needed to be expanded each time however this is not always the case and when it's not the next value was inserted in the wrong place and a value from another Seq leaked through.


louthy:
Thanks Tom (y)

PR #638, opened 2019-09-24
Implement better hash algorithm.

### LINKED ISSUE #560: Record<A> GetHashCode calculating wrong values
GetHashCode for record type is not considering the 'name' of the property while calculating hashcodes so if two properties within a record have same type, the hashcode is same for the record.

example:

```C#        
        public class TestRecord : Record<TestRecord>
        {
            public readonly Option<int> Opt1;
            public readonly Option<int> Opt2;

            public TestRecord(Option<int> opt1, Option<int> opt2)
            {
                Opt1 = opt1;
                Opt2 = opt2;
            }
        }
        static void Main(string[] args)
        {
            var testRecord1 = new TestRecord(Option<int>.None, 2);
            var testRecord2 = new TestRecord(2, Option<int>.None);

            var isHashCodeEqual = testRecord1.GetHashCode() == testRecord2.GetHashCode(); //TRUE: expected to be false

            testRecord1 = new TestRecord(Option<int>.None, Option<int>.None);
            testRecord2 = new TestRecord(1, 1);

            isHashCodeEqual = testRecord1.GetHashCode() == testRecord2.GetHashCode(); //TRUE: expected to be FALSE
        }
```

alaendle:
This should fix #560.

### DIFF
--- a/LanguageExt.Core/Utility/IL.cs
+++ b/LanguageExt.Core/Utility/IL.cs
@@ -373,8 +373,8 @@ namespace LanguageExt
                 typeof(NonRecordAttribute));
 
             var self = Expression.Parameter(typeof(A));
-            var hash = Expression.Constant(-2128831035);
-            var add = Expression.Constant(16777619);
+            var FNV_offset_basis = Expression.Constant(0x811c9dc5);
+            var FNV_prime = Expression.Constant(16777619);
             var zero = Expression.Constant(0);
 
             var Null = Expression.Constant(null, typeof(A));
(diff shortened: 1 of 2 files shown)


louthy:
Thank you, I’ll review this over the next few days

bender2k14:
Great job @alaendle!  Adding the link is very helpful.  It seems funny to me that the existing hash algorithm was already using the (smallest) 32-bit FNV prime but didn't implement (either of the) FNV hash algorithms.

I read the code.  I think it looks good.

I just have one suggestion: say somewhere that this is the implementation for 32-bit integers.  Maybe by adding a comment above `FNV_offset_basis` and `FNV_prime` since it is only these values that depend on the number of bits in the integer used.

alaendle:
I just added added the comment and also unified the types of the constants - now both are signed integers. I think this is more straight forward - since the return value of GetHashCode is defined as a signed 32-bit value by the .net framework.

bender2k14:
Computationally more straight forward, but it took me a while to see why you are so explicit about the parameters being _signed_.  The `FNV_offset_basis` as an unsigned 32-bit integer is `2166136261`, but that is slightly larger than the signed 32-bit maximum value `int.MaxValue = 2147483647`, so it "overflows" to `-2128831035`.

I think the "correct fix" is not in this code but in the Wikipedia page.  I will try to add the signed variants to Wikipedia.

bender2k14:
This thinking about overflow reminds me about something else.  This hashing code should (when written in C#) be within an [unchecked block](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/unchecked).

Is this hashing code unchecked?

IIRC, checking for over- and underflows is disabled by default, so to testing this requires not only computing a hash that would over-/underflow but also enabling such exception throwing during complication.  This is (potentially) a separate issue though, so this PR doesn't need to be held up for this reason.

PR #643, opened 2019-10-01
manyn0 parser now works correct for n==0

StefanBertels:
I think n==0 is a valid input for `manyn0`. This change fixes it.

I copied the logic from `counti` implementation (i.e. negative values are ok, same result like 0).
Other parsers seem to have different results for out of range input values, e.g. `manyn1`:

``` 
[Fact]
        public void ParseN1TimesOutOfRange()
        {
            Assert.Equal("", parse(asString(manyn1(digit, 0)), "1").ToEither().IfLeft("x"));
            Assert.Equal("", parse(asString(manyn1(digit, 0)), "").ToEither().IfLeft("x"));
        }
```

Question: Should out of range input (here: negative `n`) result in error or should it work as if minimal viable input is given or should it produce empty result? (I'm ok with "undefined" for now...)

### DIFF
--- a/LanguageExt.Parsec/Parsers/Prim.cs
+++ b/LanguageExt.Parsec/Parsers/Prim.cs
@@ -398,49 +398,52 @@ namespace LanguageExt.Parsec
         /// Enumerable of the returned values of p.
         /// </returns>
         public static Parser<Seq<T>> manyn0<T>(Parser<T> p, int n) =>
-            inp =>
-            {
-                var current = inp;
-                var results = new List<T>();
-                ParserError error = null;
-
-                int count = 0;
-
-                while (true)
+            n <= 0 
+                ? result(Seq<T>.Empty)
+                : inp =>
                 {
-                    var t = p(current);
-                    count++;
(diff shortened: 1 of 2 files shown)


louthy:
I’m away for a few days, I’ll check this when I return

louthy:
@StefanBertels Hi Stefan, do you mind adding a few more unit tests for the case when there are items to parse?

StefanBertels:
There are test cases in `ParseN0Times()`: https://github.com/louthy/language-ext/blob/5d27789adaae29e2dc04fcf1a6d9e29f3359fd71/LanguageExt.Tests/ParsecTests.cs#L532-L550

If you think there is still something missing, could you give a hint what's missing?

BTW: Maybe we could use property based testing (basically randomized tests with invariants) using FsCheck to harden the implementation, especially for edge cases. 
E.g. the output of those `manyXY` should be equal for many inputs. `either(count(n, x), manyn1(x, n - 1))` should be equivalent to `manyn1(x, n)`. Let me know if you like this (and accept additional dependencies).

PR #644, opened 2019-10-01
adds consume-only version of Iter extension

StefanBertels:
This is a very minor feature request (with code + tests).

Background I often have LINQ expressions resulting in `IEnumerable<T>` with side-effects (legacy code). I need to just execute it (no return value).

This is similar to e.g. `Consume` extension function in MoreLinq but returns `Unit`.

This new `.Iter()` version could be replaced by existing `.Iter(dummy => {})` but the latter is IMHO an ugly workaround.

I don't expect this commit to be merged -- it's just a feature request.


louthy:
I think if it was called Consume rather than Iter then I’d accept this.  We have something similar in Strict for other types, but the assumption there is that the result is cached, so I think Consume sounds reasonable

louthy:
@StefanBertels Thanks Stefan 👍

Issue #550, opened 2019-02-08
Overload for IfNone to accept an action


B-Reif:
Currently I'm using `Option`'s `IfSome` action overload to execute side effects, e.g.
```cs
optional.IfSome(value => DoImpureThings(value));
```

I noticed that `IfNone` doesn't have a similar signature that takes an `Action`.
```cs
optional.IfNone(() => DoImpureThings()); // <-- interpreted as a Func<T>, returns null, throws
```

To achieve the same thing I'm using the unsafe signature and returning null explicitly.

```cs
optional.IfNoneUnsafe(() => {
  DoImpureThings();
  return null;
});
```

My expectation was that the `IfNone` API would match the `IfSome` API. Should `IfNone` have an `Action` overload? Is this even the idiomatic way to do a side effect based on an `Option`?

bender2k14:
> My expectation was that the `IfNone` API would match the `IfSome` API.

I have also been surprised by this inconsistency.

> To achieve the same thing I'm using the unsafe signature and returning null explicitly.

You don't have to return `null` though.  You can return `Unit`, which is one of the ways in FP to indicate that a function has a side effect.

B-Reif:
Trying to do that gives me a type error; the signature expects a return value of a `T`.

bender2k14:
Oh, I see.  Then your approach wouldn't work with a value type.

In general, I think `ma.Match(Some, () => { DoImpureThings(); return None; })` would work.  I am just trying to think of a way to avoid using `null`.

B-Reif:
I mean, I'm deliberately not interested in having any kind of return value in this case.

bender2k14:
How about `ma.Match(_ => unit, () => { DoImpureThings(); return unit; })` ?

PR #653, opened 2019-10-24
Should solve request #550 (IfNone(Action) sideeffect)

StefanBertels:
see https://github.com/louthy/language-ext/issues/550

I personally would avoid this style but I see no point in having `IfSome(Action)` without having `IfNone(Action)`.


louthy:
Thanks @StefanBertels

PR #654, opened 2019-10-24
make TraversAsync unit test pass with corecount < 6

StefanBertels:
I increased timeout slightly to make the test pass more stable.


louthy:
Thanks Stefan!

Issue #665, opened 2019-11-21
Discriminated [Union] Code generation issues


StefanBertels:
Hi,

I just checked the latest and greatest [Union] feature described here:
https://github.com/louthy/language-ext/releases/tag/3.3.34

I used v3.3.39.

Two things I stumbled upon:

1. If there are some imports (`using`) missing, this will raise errors. Probably generated code should add them or even better use full type names.

2. Defining Unions within a class (nested class) does not work without making the outer class `partial`.

louthy:
> If there are some imports (using) missing, this will raise errors. Probably generated code should add them or even better use full type names.

Fix on its way.

> Defining Unions within a class (nested class) does not work without making the outer class partial.

This isn't supported behaviour, but it seems the solution of making the outer class partial is there if you need it.  I can't think of another way that _could_ work without the outer class being partial.

louthy:
Fixed in `3.3.40`

PR #668, opened 2019-11-22
Use full name for EqDefault and OrdDefault in code gen

nekronos:
This fixes first point in https://github.com/louthy/language-ext/issues/665. Not sure if this is the correct fix, maybe we want the appropriate `using` to be added instead? So feel free to reject this PR.

The issue can be reproduced by adding a new file in the TestBed project with:
```c#
using LanguageExt;
using static LanguageExt.Prelude;

namespace TestBed
{
    [Union]
    public abstract partial class Foo<T>
    {
        public abstract Foo<T> Bar(T x);
        public abstract Foo<T> Baz(T x);
    }
}
```

### DIFF
--- a/LanguageExt.CodeGen/CodeGenUtil.cs
+++ b/LanguageExt.CodeGen/CodeGenUtil.cs
@@ -778,7 +778,7 @@ namespace LanguageExt.CodeGen
                                                             SyntaxKind.SimpleMemberAccessExpression,
                                                             DefaultExpression(
                                                                 GenericName(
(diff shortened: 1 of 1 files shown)


louthy:
Thanks for raising this, but the solution isn't quite as simple, it needs to be a qualified name, not an identifier.  I have a fix and will be deploying it later, so will close this PR down.

louthy:
@nekronos  Fixed in `3.3.40`

nekronos:
@louthy Great, thanks! 👍

Issue #669, opened 2019-11-26
no parseEnum wrapper for "ignoreCase" version of Enum.TryParse


jltrem:
There is no parseEnum wrapper for `Enum.TryParse(string value, bool ignoreCase, out TEnum result)`.

I'll submit a PR shortly

PR #670, opened 2019-11-26
Add Prelude.parseEnumIgnoreCase<TEnum>(string value)

jltrem:
This addresses issue #669

### DIFF
--- a/LanguageExt.Core/Prelude/Prelude_Parse.cs
+++ b/LanguageExt.Core/Prelude/Prelude_Parse.cs
@@ -111,14 +111,26 @@ namespace LanguageExt
             where TEnum : struct =>
             Parse<TEnum>(Enum.TryParse, value);
 
-[Pure]
-public static Option<IPAddress> parseIPAddress(string value) =>
+        [Pure]
+        public static Option<TEnum> parseEnumIgnoreCase<TEnum>(string value)
+            where TEnum : struct =>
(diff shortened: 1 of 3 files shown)


louthy:
btw, I haven't forgotten about this, I'm just not 100% on the need for separate functions that capture the ignore-case flag, and right now it's not top of my mind to care too much about it, so I'm pressing pause on this PR until I think about it fully.

jltrem:
thanks, Paul.  It seemed a bit dubious to me too.  FWIW the request came from someone on my team implementing config file parsing. (He hadn't been keen on using language-ext until recently, so I was pleased he brought it up.)  If you think of a better approach I'd be happy to rework it.

louthy:
@jltrem Hi Joe, I think I'd prefer there to just be one `parseEnum` with an `ignoreCase` boolean, like the original.

jltrem:
Not sure I follow you on "like the original".  Do you mean change the existing `public static Option<TEnum> parseEnum<TEnum>(string value)` to have an ignoreCase boolean?  If so, would you want it to have a default `false` to avoid a breaking change?

louthy:
Sorry to be unclear, I forgot the `parseEnum` already existed.  I think probably have.  I'm changing my mind now, because there's always value to single argument functions in functional programming.  So, I think I'll accept as is

louthy:
Thanks @jltrem

PR #680, opened 2019-12-18
adds failing test for Union serialization

StefanBertels:
Some recent update made my serialization fail. This test shows why.

IMHO serialization should work when serialized property name has same case as code property, i.e. first letter upper case.

BTW:  I have a custom converter for Union itself which requires that I add two custom attributes to be able to detect via reflection whether a type is Union base type (Shape) and whether a type is one of those variants (Circle). Example code following in a separate issue.

### DIFF
--- a/LanguageExt.Tests/UnionTests.cs
+++ b/LanguageExt.Tests/UnionTests.cs
@@ -1,0 +1,23 @@
+using Newtonsoft.Json;
+using Xunit;
+
+namespace LanguageExt.Tests
+{
+    public partial class UnionTests
+    {
+        [Union]
+        public abstract partial class LightControl
+        {
+            public abstract LightControl OnOff(bool enabled);
+            public abstract LightControl Dimmer(int value);
+        }
+
+        [Fact]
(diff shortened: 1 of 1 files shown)


StefanBertels:
The serialization constructor probably should be fixed (case), but I think there is another related problem.

We can tell Newtonsoft.Json to not use ISerializable this way: 

```
new JsonSerializerSettings()
{ 
    ContractResolver =  new DefaultContractResolver(){IgnoreSerializableInterface = true, IgnoreIsSpecifiedMembers = true}
}
```
This does not work either because Newtonsoft.Json somehow does not find the regular constructor (but it seems to work if the serialization constructor is removed).

StefanBertels:
According to
https://docs.microsoft.com/en-us/visualstudio/code-quality/ca2229?view=vs-2019
the serialization constructor should be private (resp. protected) which will solve the issue with the regular constructor not being found.

StefanBertels:
Sorry, accidentally requested master for merge. Fixed in #682.

Issue #681, opened 2019-12-18
Union serialization: How to make this less hacky?


StefanBertels:
Related to #680.

I build a little workaround (below) for safe union (de)serialization. It is safe regarding type because the type name is stored explicitly. 

```
[Union]
[MyUnionRead]
[MyUnionWrite]
abstract partial class UnionTestClass
{
	public abstract UnionTestClass A();
	public abstract UnionTestClass B(int i);
	public abstract UnionTestClass C(int i);
	public abstract UnionTestClass D(int i, int j);
}

[AttributeUsage(AttributeTargets.Class, Inherited = false)]
public class MyUnionReadAttribute : Attribute
{
}

[AttributeUsage(AttributeTargets.Class, Inherited = true)]
public class MyUnionWriteAttribute : Attribute
{
}

public class UnionJsonReadConverter : JsonConverter
{
	public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer) => throw new NotImplementedException();

	public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
	{
		string typeName = null;
		JObject typeValueToken = null;
		if (reader.TokenType != JsonToken.StartObject) throw new Exception();
		while (reader.Read())
		{
			if (reader.TokenType == JsonToken.EndObject) break;
			if (reader.TokenType != JsonToken.PropertyName) throw new Exception();
			if ((string) reader.Value == "Type")
			{
				if (!reader.Read()) throw new Exception();
				if (typeName != null) throw new Exception();
				if (reader.TokenType != JsonToken.String) throw new Exception();
				typeName = (string) reader.Value;
			}
			else if ((string) reader.Value == "Value")
			{
				if (typeValueToken != null) throw new Exception();
				if (!reader.Read()) throw new Exception();
				if (reader.TokenType != JsonToken.StartObject) throw new Exception();
				typeValueToken = JObject.Load(reader);
			}
		}

		var type = objectType.Assembly.GetTypes().Where(_ => objectType.IsAssignableFrom(_) && _.Name == typeName).Single();
		var result = typeValueToken.ToObject(type, serializer);
		return result;
	}

	public override bool CanWrite => false;
	public override bool CanConvert(Type objectType) => objectType.IsClass && objectType.GetCustomAttribute<MyUnionReadAttribute>() != null;
}
public class UnionJsonWriteConverter : JsonConverter
{
[...]

PR #682, opened 2019-12-18
Fix union deserialisation

StefanBertels:
Some recent update made my serialization fail. This test shows why.

IMHO serialization should work when serialized property name has same case as code property, i.e. first letter upper case.

The serialization constructor probably should be fixed (case), but I think there is another related problem.

We can tell Newtonsoft.Json to not use ISerializable this way: 

```
new JsonSerializerSettings()
{ 
    ContractResolver =  new DefaultContractResolver(){IgnoreSerializableInterface = true, IgnoreIsSpecifiedMembers = true}
}
```
This does not work either because Newtonsoft.Json can't identify the regular constructor. According to
https://docs.microsoft.com/en-us/visualstudio/code-quality/ca2229?view=vs-2019
the serialization constructor should be private (resp. protected) which will solve the issue with the regular constructor not being found.

https://github.com/louthy/language-ext/blob/master/LanguageExt.CodeGen/CodeGenUtil.cs#L1983

BTW: related issue #681 -- how to make Union serialization work.

### DIFF
--- a/LanguageExt.CodeGen/CodeGenUtil.cs
+++ b/LanguageExt.CodeGen/CodeGenUtil.cs
@@ -1980,7 +1980,7 @@ namespace LanguageExt.CodeGen
                             Identifier(typeName))
                         .WithModifiers(
                             TokenList(
-                                Token(SyntaxKind.PublicKeyword)))
+                                Token(SyntaxKind.ProtectedKeyword)))
                         .WithParameterList(
                             ParameterList(
                                 SeparatedList<ParameterSyntax>(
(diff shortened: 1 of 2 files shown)


StefanBertels:
Codegen 3.3.40 seems to be the last "good" version.

BTW: it was quite hard to locate/fix the problem. I finally debugged Newtonsoft.Json etc.
The quick fix "downgrade LanguageExt" did not work well because the code wasn't automatically re-generated when building. I had to touch the source file / make a clean build. Is there a solution to this? Code generation should re-generate everything if codegen package is updated/downgraded without manual cleanup steps. TIL: even my unit test script needs a "make clean" before...

louthy:
Well spotted!  Thanks @StefanBertels

PR #685, opened 2019-12-20
Union serialization issue (part 2)

StefanBertels:
There are some more issues with serialization of union types. I tried to find a fix but seems it isn't that easy and it depends on some design decisions. This pull request just shows some issue!

IMO serialization should work with Newtonsoft.Json without doing too much custom magic. Personally would take the extra step with custom converters / contract resolvers but that might be a burden for users in general.

I see why you want support `ISerializable` and the basic idea is great. But this probably won't work together with customization (like `JsonConverter` / `JsonContractResolver` ...). I use custom code e.g. for (de)serializing `Option` which is compatible to `Nullable`.

1. Here I changed `protected` to `private` for serialization constructor (compare #682) as we get compilation errors for structs otherwise. But this must be changed to be `protected` for `abstract class` to enable inheritance.

2. I removed `ISerializable` for now to make Newtonsoft.Json not use this interface by default -- instead Newtonsoft.Json will use it's own (de)serializer / custom (de)serializer which means regular constructor. This is the way `Record<>` works now. `Record` base class now implements `GetObjectData` and serialization constructor but this only get's enabled / visible if we mark our custom record class with `ISerializable` (and implement the serialization constructor by calling `base`). At least this works fine with records and allow both ways.

Currently this style would not work for Union types because it's not allowed to add `ISerializable` in my custom code to my `[Union]` marked (abstract) class. But hopefully CodeGenUtil could take care of this.

3. The included tests in `UnionJsonSerializerTests`show simple serialization of Union instance classes. IMHO this is what I would expect to work out of the box. One test fails due to `Tag` property. Maybe this should be a method. If you think this will be helpful for detecting the type (for deserialization) read below.

4. The included tests in `UnionCustomJsonSerializerTests` show how custom serialization could work. I could remove one custom attribute (compare #681). Would be nice if there would be an attribute-less way to detect union type base classes. Somehow I cannot check for `UnionAttribute` (invisible to me).

5. Regarding `Tag` and deserialization. As you can see in my Custom serializer I took another approach. I think this is cleaner. Newtonsoft.Json has as hacky way of interface deserialization including some `$type` property in json. It's hard to write custom code to read this because you have to read details of the object to know the type of the object. Even if you get this working (similiar to Newtonsoft.Json) this will result in problems in practice (even if `Tag` would be an `enum` / string of the type name). I think the cleaner way is to have a nested `JObject` as in my custom serializer.


louthy:
Thanks @StefanBertels !

PR #693, opened 2020-01-08
EitherData Left status from IsRight to IsLeft

mcurtois:
Noticed this when doing TryAsync(_Task in failed state_).ToEither(). The result was always in the Right state.
As a side note, I'd like to say how much I'm enjoying using this library. It's made programming in C# fun again. Keep up the good work!


louthy:
Good catch, thanks!

PR #700, opened 2020-01-17
Bug fix for #699 where manyn parses short strings

### LINKED ISSUE #699: manyn is accepting fewer than expected matches if more than 1 away from n
I've found that manyn isn't failing if the input string isn't long enough to reach n by more than 1 away.

I added this test to ParsecTests and it fails:

```
[Fact]
public void ParseNTimesFailIfMoreThan1Away()
{
    var p = asString(manyn(digit, 4));

    var r = parse(p, "12").ToEither();
    Assert.True(r.IsLeft);
}
```

I believe it's all down to this line, although I can't understand why it's doing what it's doing to give any more pointers.

```
// eerr
return count == n
  ? ConsumedError<Seq<T>>(mergeError(error, t.Reply.Error))
  : EmptyOK(Seq(results), current, mergeError(error, t.Reply.Error));
```

P.S. Great library, thank you

bender2k14:
Fixes #699 

I arrived a this change (in the [second commit](https://github.com/louthy/language-ext/commit/4eda0278dd57a955d146db4c6d6c1e853bd5645c)) by trial-and-error with good guessing.  Or, said another way, I didn't really know what I was doing, but I found this solution that seems to work.

In particular, the parsing error message for the test case @Jmaharman gave in #699 (when trying to parse 4 digits but only given "12") is
> error at (line 1, column 3): unexpected end of stream, expecting digit

I think that is the correct error message.

The code I deleted allowed the `manyn` parser to return after processing fewer than `n` things.  I think this bug is partially due to copying and pasting the nearly identical code for the `many` and `manyn0` parsers, which do not have a minimum number of items they need to parse in order to return success.

The thing I understood the least was what error message to return.  I knew I wanted to return a parser error created from the function `EmptyError`, but I am totally guessing that the parser error to pass into that call should be `mergeError(error, t.Reply.Error)`.  It does seem like a good guess though, because the example error message that I shared above looks good to me.

### DIFF
--- a/LanguageExt.Parsec/Parsers/Prim.cs
+++ b/LanguageExt.Parsec/Parsers/Prim.cs
@@ -429,7 +429,7 @@ namespace LanguageExt.Parsec
                         if (t.Tag == ResultTag.Empty && t.Reply.Tag == ReplyTag.OK)
                         {
                             // eok, eerr
-                            return EmptyError<Seq<T>>(new ParserError(ParserErrorTag.SysUnexpect, current.Pos, "many: combinator 'manyn' is applied to a parser that accepts an empty string.", List.empty<string>()));
+                            return EmptyError<Seq<T>>(new ParserError(ParserErrorTag.SysUnexpect, current.Pos, "many: combinator 'manyn0' is applied to a parser that accepts an empty string.", List.empty<string>()));
                         }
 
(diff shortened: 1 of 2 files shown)


louthy:
Thanks @bender2k14 👍

PR #703, opened 2020-01-24
Fix Try.Filter / Try.BiFilter

StefanBertels:
1. `Filter` of `Try` should not execute `pred` if in faulted state.

2. `BiFilter` of `Try` should probably not be allowed to switch from faulted state back to success. This was the case when the `Fail` predicate returned `true`. `res.Value` should not be used when in faulted state and I stumbled upon getting success state with a default value. My fix would mean that `Fail` predicate can just decide whether to stay in faulted state with original exception (Fail pred => true) or to go to bottom state (Fail pred => false). I compared with `Either` but seems there is no `bifilter`. I personally wouldn't mind to remove `Try.BiFilter`.


louthy:
Thanks @StefanBertels 👍

PR #714, opened 2020-02-11
Additional New method fails

StefanBertels:
Minor issue (if any), just stumbled upon this.

The test below fails. I you add `@` in line 14 before `New` (method name `@New`) it works.
This PR is based on commit 88600b71 (newer commits don't compile at my machine).

### DIFF
--- a/LanguageExt.Tests/RecordCodeGenTests.cs
+++ b/LanguageExt.Tests/RecordCodeGenTests.cs
@@ -1,0 +1,25 @@
+using System;
+using Xunit;
+
+namespace LanguageExt.Tests
+{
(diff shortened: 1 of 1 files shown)


louthy:
That's actually a feature, but you can get around it by doing this:
```c#
    [Record]
    public partial class MyRecord
    {
        public int DateId { get; }
    }

    public partial class MyRecord
    {
        public static MyRecord New(DateTime dt) => MyRecord.New(dt.Year * 10000 + dt.Month * 100 + dt.Day);
    }
```
Of course it would be preferable if `New` methods that didn't have the same types as the one provided were allowed, but this will continue to work if that is solved.

PR #720, opened 2020-02-17
make Ord<T> usable as IComparer<T>

StefanBertels:
I don't know whether this is a good or bad thing.

After trying to use `Ord` with other code (e.g. LINQ OrderBy) I see some demand for conversion to types like IComparer.

### DIFF
--- a/LanguageExt.Core/TypeClasses/Ord/Ord.cs
+++ b/LanguageExt.Core/TypeClasses/Ord/Ord.cs
@@ -4,7 +4,7 @@ using System.Diagnostics.Contracts;
 namespace LanguageExt.TypeClasses
 {
     [Typeclass("Ord*")]
-    public interface Ord<A> : Eq<A>
+    public interface Ord<A> : Eq<A>, System.Collections.Generic.IComparer<A>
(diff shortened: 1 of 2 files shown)


StefanBertels:
Maybe `OrdReverse` in  https://github.com/louthy/language-ext/pull/720/commits/8ebcd48a1a56f1f717472f46b6e1c62bf1ddbb14 might be useful as public class in LanguageExt.Core.

StefanBertels:
BTW: `OrdDescending` might be another name for `OrdReverse`.

iamim:
I actually had a couple of cases where I really wanted this to be possible.

louthy:
`OrdDesc` would be my preference.

louthy:
Thanks Stefan 👍

StefanBertels:
Hi, seems this feature got lost.

louthy:
It breaks the class resolution system.  You can call `.ToComparable()`

https://github.com/louthy/language-ext/blob/main/LanguageExt.Core/TypeClasses/Ord/Ord.cs#L39

PR #726, opened 2020-02-27
Fix Seq.TakeWhile inserting 4 empty elements to the resulting sequence

LAJW:
- First commit contains the minimal fix, as I don't quite know what strictseq class is for and why does it have to have those empty elements in the front (I don't think it has to).
 - Second commit contains some tests for this function with an example which lead me to discover this bug.
 - Third commit contains proposed refactor/optimization - I don't think takeWhile should buffer the elements. Lazy sequence is the way to go.

### DIFF
--- a/LanguageExt.Core/DataTypes/Seq/Seq.cs
+++ b/LanguageExt.Core/DataTypes/Seq/Seq.cs
@@ -26,9 +26,6 @@ namespace LanguageExt
 #pragma warning restore CS0618
         IComparable<Seq<A>>, IEquatable<Seq<A>>
     {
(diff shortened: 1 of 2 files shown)


louthy:
The reason for the space is for the sequence to grow into (via `Cons`).  It's an optimisation [that has serious performance benefits](https://github.com/louthy/language-ext/blob/master/Performance.md).  It is unclear to me if you're fixing a bug here, or have decided that you don't like the implementation?

louthy:
@LAJW Hi Lukasz, I've taken a slightly different route to fixing it, thanks for bringing it up thought.  It's deployed in `3.4.12`

PR #729, opened 2020-03-07
seq traverse unit tests

harrhp:
https://github.com/louthy/language-ext/projects/1#card-34099493

### DIFF
--- a/LanguageExt.Tests/Transformer/Traverse/SeqT/Collections/Arr.cs
+++ b/LanguageExt.Tests/Transformer/Traverse/SeqT/Collections/Arr.cs
@@ -1,0 +1,28 @@
+using Xunit;
+using static LanguageExt.Prelude;
+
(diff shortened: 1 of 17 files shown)


louthy:
Looks great, thanks @harrhp !  

There's a couple of use-cases missing if you have time?

[Take a look at the `HashSet` tests I just added](https://github.com/louthy/language-ext/blob/hand-written-traverse/LanguageExt.Tests/Transformer/Traverse/HashSet/Collections/Arr.cs), the 'cross product' (`ArrHashSetCrossProduct`) and mixture of empty and non-empty (`ArrOfEmptiesAndNonEmptiesIsEmpty`), the results should be the same for `Seq` too. 



If you don't have time, it's not a massive issue as the code behind all this is the same, so I know it works.  Your help so far is very much appreciated 👍

PR #733, opened 2020-03-08
make seq traverse tests consistent with other tests

harrhp:
continuation of https://github.com/louthy/language-ext/pull/729


louthy:
Thanks 👍

Issue #525, opened 2018-11-24
TryAsnyc with non generic task


eisendle:
Hi Paul,

I accidentally used TryAsync with a non generic Task. When calling TryAsync.Match, I expected that the task is being awaited, but as there is no overload for TryAync with Task, instead this overload was used:

`public static TryAsync<A> TryAsync<A>(A v)`

As the success function of match didn´t use the result (in this case Task), this misbehaviour was hard to spot.

In order to make TryAsync work with Task correctly and prevent other users to fall into this pifall, I´d like to propose two new overloads for TryAsync:

` public static TryAsync<Unit> TryAsync(Task v))`

 `public static TryAsync<Unit> TryAsync(Func<Task> f)`

I´m happy to supply a pull request, if you agree to add the overloads

bender2k14:
`Task` is the asynchronous equivalent of`void`.  Neither are consistent with functional style..

In my local project, I have created an extension method that lifts `Task` to `Task<Unit>`.  I prefer adding this one method than adding overloads everywhere for `Task`.

WarrenFerrell:
Could a funasync method be added to Prelude so the lift function is already provided.

louthy:
@WarrenFerrell `task.ToUnit()` already exists, which then gives all the other fuctionality.

WarrenFerrell:
@louthy Thank you, I figured there had to be something but this post was the only thing I was able to find... I added a PR #741 since a significant number of client libraries are deprecating their synchronous apis and this is likely to be an issue going forward.

PR #741, opened 2020-03-12
#525 add explanation for making void functions return a value.

WarrenFerrell:
Took slightly longer than I'd expect to figure out the preferred method to functionalize void and Task functions. New to functional programming so won't be offended if there is actually a better method. Included both returning the function and writing the function because many people reading the readme are also likely new to functional programming and might not be completely used to the idea of functions as values.


louthy:
Thanks Warren 👍

PR #745, opened 2020-03-22
Use value-typed enumerators

aalmada:
The collections in `System.Collections.Generic` use a trick to improve performance of enumeration. The [public `GetEnumerator()` method returns a value-type](https://github.com/dotnet/runtime/blob/ff8e91121ced72b72daddee2a4c9b16679223fb4/src/libraries/System.Private.CoreLib/src/System/Collections/Generic/List.cs#L576) instead of the `IEnumerator<T>` interface. Calls to its methods in value-types are not virtual, which are much more performant. Also, these enumerators are not allocated on the heap.

One other trick used in `System.Collections.Immutable` is to the [avoid implementation of `IEnumerator<T>` by the value-type whenever possible](https://github.com/dotnet/runtime/blob/ff8e91121ced72b72daddee2a4c9b16679223fb4/src/libraries/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableStack_1.Enumerator.cs#L16). Implement only the property `Current` and the method `MoveNext()`. This is supported by `foreach` and does not force it to use a `try`/`catch` clause, allowing it to be inlinable.

This PR changes the enumerators to be value-types. It also changes the return types of the public `GetEnumerator()` methods so that the enumerators are not boxed.

For this, it is required that the enumerator is a public type. This PR changes the enumerator types to `public` but its constructors to `internal` so that instances can only be created by this library. This PR changes `ListEnumerator<>` to be an outer type so that no changes have to be made to `ListModule`.

Enumerator that implement `IEnumerator<T>` should be disposed so, this PR adds the `using` keyword wherever necessary.

While comparing arrays, it's good to avoid the use of enumerators. The indexer is much more performant. This PR uses [`IStructuralEquatable.Equals()`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.istructuralequatable.equals?view=netcore-3.1) and [`IStructuralComparable.CompareTo`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.istructuralcomparable.compareto?view=netcore-3.1) to compare arrays.


aalmada:
The enumerators implemented using `yield` are reference-types. Changing these would require implementing complex enumerators which require more testing and are harder to maintain. Something to consider at a later time.

aalmada:
@louthy I reverted all the changes to Equals and Compare that you mentioned

louthy:
@aalmada Thanks! 👍

PR #751, opened 2020-04-01
Fix Union CompareTo causing Stack Overflow

LAJW:
Generated code looked like:
```csharp
   class Derived
        public int CompareTo(object obj) => obj is Base p ? CompareTo(p) : 1;
```
There's no downcast and so the function ended up calling itself indefinitely.

This popped up when I tried calling `Assert.Equal` on two Union instances - it uses `.CompareTo` in XUnit

`.Equals` method didn't have that problem, but I included some unit tests anyway.


louthy:
Thanks Lukasz

PR #754, opened 2020-04-06
Update CodeGen pacakge to use v0.7 of CodeGeneration.Roslyn

garyng:
- [x] move attributes out into separate project (as per [readme](https://github.com/AArnott/CodeGeneration.Roslyn#separate-out-the-attribute))
- [x] move generators out into separate project (now the `CodeGenerationAttribute` uses string instead of with `typeof`)
- [x] add samples - a cli using the codegen, and a netstandard library using the codegen (made possible by making changes to the reference, as per [readme](https://github.com/AArnott/CodeGeneration.Roslyn#customize-generator-reference))
- [x] fix `TestBed` sample
- [x] test all generators (only tested with `Union` and `Record`)
- [x] create [metapackage](https://github.com/AArnott/CodeGeneration.Roslyn#create-the-metapackage), seems like [CodeGeneration.Roslyn.PluginMetapackage.Sdk](https://www.nuget.org/packages/CodeGeneration.Roslyn.PluginMetapackage.Sdk) is [added](https://github.com/AArnott/CodeGeneration.Roslyn/pull/205) but not pushed to nuget yet
- [ ] update [wiki on codegen](https://github.com/louthy/language-ext/wiki/Code-generation#setup)

_Related to https://github.com/AArnott/CodeGeneration.Roslyn/issues/191, https://github.com/AArnott/CodeGeneration.Roslyn/issues/193_

~I'm not sure whether separating the project into a metapackage, generators, and attributes is necessary though. After some experimentation, the following usage does not work:~
~1. a library that only references the `Attributes` package
2. a console app that reference the library, the `Generators` and `Attributes`, and `CodeGeneration.Roslyn.Tool`.~
~The codegen will not run on the library unless it also references the `Generators` and `CodeGeneration.Roslyn.Tool`.~
see https://github.com/louthy/language-ext/pull/754#issuecomment-610246410


amis92:
> After some experimentation, the following usage does not work

How does it "not work"? I want to help.

garyng:
> 
> 
> > After some experimentation, the following usage does not work
> 
> How does it "not work"? I want to help.

Uhmm, let me elaborate:

Lets say there is `Attributes` and a `Generators` package; where `Attributes` define all the attributes that is used for annotating classes/structs/etc.; and the `Generators` contains all the codegen code

As a consumer, 
1. I create a new netstandard library, `Models`, only uses the attributes defined in the `Attributes` to annotate my classes/etc.
1. I then create a netcoreapp, `App` -- this reference the `Attributes`, `Generators`, `CodeGeneration.Roslyn.Tool`, and the netstandard library `Models`.

I first though that compiling `App` would be sufficient for the tool to discover and run codegen on all the annotated classes from `Models`; but it turns out it doesn't -- the `Models` library needs to reference `Generators` and `CodeGeneration.Roslyn.Tool` as well.

Adding those references (especially `Generators`) requires a lot of manual work, so I was hoping:
1. the metapackage would solve this problem, so the `Models` library can reference the metapackage and everything works
2. or the tool will automatically discover all stuff annotated with `Attributes` and run codegen for them accordingly

amis92:
Ok, so:
1. codegen is only working on source files in the current project (so no generation for Models in App), as you've found out - it's very much by design
2. Metapackage solves this problem: you only add single reference to it in the consuming project (Models); and everything should work perfectly.

Note that there's a major difference between using codegen P2P (ProjectReferefence) and as NuGets (PackageReference).

Everything should be documented in CodeGeneration.Roslyn's Readme. :)

garyng:
As a library consumer, option 2 is the way to go right? The library author create a metapackage, the users just use the metapackage?

> Note that there's a major difference between using codegen P2P (ProjectReferefence) and as NuGets (PackageReference).

What you meant is directly referencing the `Generators` project (ie. they are in the same solution) is different from consuming it as a nuget package (eg. the metapackage), right?

garyng:
As stated in the readme (and in https://github.com/AArnott/CodeGeneration.Roslyn/issues/191#issuecomment-610869820), the metapackage is only applicable for nuget consumer. When the generator lives in the same solution as the consumer, some modifications need to be done manually to the `ProjectReference` (see https://github.com/AArnott/CodeGeneration.Roslyn#customize-generator-reference).

---

Lots of thanks to @amis92 for helping me!

LanguageExt.CodeGen/LanguageExt.CodeGen.csproj
```diff
@@ -22,26 +17,15 @@
     <PackageLicenseUrl>https://github.com/louthy/language-ext/blob/master/LICENSE.md</PackageLicenseUrl>
     <EnableDefaultCompileItems>false</EnableDefaultCompileItems>
     <DocumentationFile></DocumentationFile>
-    <OutputType>library</OutputType>
     <AssemblyVersion>3.0.0.0</AssemblyVersion>
     <FileVersion>3.0.0.0</FileVersion>
     <NoWarn>1701;1702;1705;IDE1006;CS1591;CS1573;CS1712;CS1570;CS1711;CS1572;CS1587</NoWarn>
-
+    <IncludeBuildOutput>false</IncludeBuildOutput>
```
amis92, line 23: In stable release, I've updated the Metapackage.Sdk to include that by default so you can remove it here :)

LanguageExt.CodeGen/LanguageExt.CodeGen.csproj
```diff
@@ -1,16 +1,11 @@
...
-  
-  <!--<Sdk Name="CodeGeneration.Roslyn.Plugin.Sdk" Version="0.7.5-alpha" />-->
+  <Sdk Name="CodeGeneration.Roslyn.PluginMetapackage.Sdk" Version="0.7.63" />  
 
   <PropertyGroup>
     <PackageVersion>3.4.13</PackageVersion>
     <PackageId>LanguageExt.CodeGen</PackageId>
     <Title>LanguageExt.CodeGen</Title>
-    <TargetFrameworks>netstandard2.0;net461</TargetFrameworks>
+    <TargetFramework>netstandard2.0</TargetFramework>
```
amis92, line 8: I think this can target `netstandard1.0`, same as Attributes package - you'll gain wider compatibility/applicability.

amis92:
Also a general note, it seems csproj files are using tab indent.

LanguageExt.CodeGen/LanguageExt.CodeGen.csproj
garyng: Sure!

LanguageExt.CodeGen/LanguageExt.CodeGen.csproj
garyng: Changed!

garyng:
> 
> 
> Also a general note, it seems csproj files are using tab indent.

Yeah, my visual studio is configured to use tabs... I will replace them with spaces for consistency.

amis92:
After this is done, can you please post an issue back in CG.R to add this in the Migration wiki as a helpful example? :)

garyng:
> 
> 
> After this is done, can you please post an issue back in CG.R to add this in the Migration wiki as a helpful example? :)

Like this? https://github.com/AArnott/CodeGeneration.Roslyn/issues/223

amis92:
@louthy If there's anything I can help with to push it, please don't hesitate to ask :) I'm eagerly awaiting wider adoption ;)

garyng:
@amis92 Seems like @louthy will be busy for these few weeks because of work, https://github.com/louthy/language-ext/issues/755

ffgiraldez:
Hi, please,I need this, I don't know why but I have both 3.1 and 2.1 runtime and sdk but vs for mac does not recognize me codegen

louthy:
@garyng I've made some pretty large changes since you started this.  I'm ready to start looking into it, but I wonder if you'd mind closing this PR down and re-doing it with the latest `main` branch?  I realise it's a bit of work, but it'll be the quickest way of getting this into the project.  

Thanks!

garyng:
@louthy Sure! I'll look into it probably on the weekends. 
Seems like Microsoft also announced [Source Generators](https://devblogs.microsoft.com/dotnet/introducing-c-source-generators/), which might possibly replace `CodeGeneration.Roslyn` (https://github.com/AArnott/CodeGeneration.Roslyn/issues/229)

louthy:
@garyng I'm going to close this down, because the next step probably will be with Source Generators.  Thanks for the effort though, sorry it didn't work out.

Issue #728, opened 2020-03-05
Help requested: unit tests for `Traverse`


louthy:
An ongoing thorn in my side has been the behaviour of `Traverse` and `Sequence` for certain pairs of monadic types (when nested).  These issues document some of the problems:

https://github.com/louthy/language-ext/issues/552
https://github.com/louthy/language-ext/issues/651

The `Traverse` and `Sequence` functions were previously auto-generated by a T4 template, because for 25 monads that's `25 * 25 * 2 = 1250` functions to write.  In practice it's a bit less than that, because not all nested monads should have a `Traverse` and `Sequence` function, but it is in the many hundreds of functions.

Because the same issue kept popping up [I decided to bite the bullet and write them all by hand](https://github.com/louthy/language-ext/tree/hand-written-traverse/LanguageExt.Core/Transformer/Traverse).  This has a number of benefits:

* The odd rules of various monads when paired can have bespoke code that makes more sense than any auto-generated T4 template could ever build.  This fixes the bugs that keep being reported and removes the surprising nature of `Traverse` and `Sequence` working most of the time, but not in all cases.
* I'm able to hand-optimise each function based on what's most efficient for the monad pairing.  This is especially powerful for working with `Traverse` and `Sequence` on list/sequence types.  The generic T4 code-gen had to create singleton sequences and the concat them, which was super inefficient and [could cause stack overflows](https://github.com/louthy/language-ext/pull/725/files).  Often now I can pre-allocate an array and use a much faster imperative implementation with sequential memory access.  Where possible I've tried to avoid nesting lambdas, again in the quest for performance but also to reduce the amount of GC objects created.  I expect a major performance boost from these changes.  
* The lazy stream types `Seq` and `IEnumerable` when paired with `async` types like `Task`, `OptionAsync`, etc. can now have bespoke behaviour to better handle the concurrency requirements (These types now have `TraverseSerial` and `SequenceSerial` which process tasks in a sequence one-at-a-time, and `TraverseParallel` and `SequenceParallel` which processes tasks in a sequence concurrently with a _window_ of running tasks - that means it's possible to stop the `Traverse` or `Sequence` operation from thrashing the scheduler.

## Help

Those are all lovely things, but the problem with writing several hundred functions manually is that there's gonna be bugs in there, especially as I've implemented them in the most imperative way I can to get the max performance out of them.  

[...]

PR #756, opened 2020-04-06
Identity traverse tests

blakeSaucier:
As mentioned in #728 - Identity traverse tests.

Looks like `Sequence` isn't defined for `Identity<A>` so all the tests use `.Traverse(identity)`

Removed some extra generic parameters in Identity.cs which confused C# a bit.


louthy:
Thanks @blakeSaucier 👍

PR #761, opened 2020-04-13
Fix bug 758

### LINKED ISSUE #758: SequenceParallel with TryAsync reduces the source collection
In 3.4.20-beta release when executed with a fake task like `Task.FromResult(i)` SequenceParallel works correctly. However, when tried with a real task like accessing a MongoDB it would only return the same number of elements as the window size.

Here is the failing unit test:

```C#
[Fact]
public async Task SequenceParallelReturnsTheOriginalNumberOfElements()
{
    var tries = from i in Range(0, 100) select TryAsync(() => Task.Delay(10).ToUnit());
    var count = await tries.SequenceParallel(10).Map(Enumerable.Count).IfFailThrow();
    Assert.Equal(100, count);
}
```

bender2k14:
Fixes #758

### DIFF
--- a/LanguageExt.Core/DataTypes/Task/Task.Extensions.cs
+++ b/LanguageExt.Core/DataTypes/Task/Task.Extensions.cs
@@ -323,14 +323,14 @@ namespace LanguageExt
                     var (s, outerTask) = GetNext();
                     if (!s) break;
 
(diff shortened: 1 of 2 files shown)


louthy:
Thanks @bender2k14 👍

PR #762, opened 2020-04-14
Add BiBind to Free Code-Gen

### LINKED ISSUE #757: Add BiBind to Free
Firstly,
@louthy I note your info issue re: availability, all good, this is a non-urgent feature suggestion.

While rolling out some features using the `Free` Monad / Interpreter pattern we have found the `Free` code-gen to be great help in reducing some of the boilerplate needed for the pattern.

However we have come across one limitation with the current implementation and that is the lack of an ability to handle or manage the failure path. More specifically the `Free` is lacking a `BiBind` function that will allow you to recover from a failure path and get back on to the success path. This also limits your ability to hook into the failure path for error logging etc.

I tried implementing with the current `Free` but realised that the resulting tree that the `Free` builds up only stores the success path by means of the `Next` field. So my idea was to add an additional `FailNext` field that will allow the Interpreter to choose between the branches based on the outcome of the current action.

I have created a sample repo https://github.com/chadedrupt/free-monad-fail-state where this idea is implemented and demonstrated.

I could be missing something here, so if anyone is aware of something I'm missing that would be great.

Also happy to knock-up a PR for the `Free` code-gen if you feel this is a feature worth implementing.

pauledrupt:
Resolves #757 
The idea is to have the Free code-gen handle the success path as well as the fail path by adding code-gen'd `BiMap` and BiBind implementations.
This adds a `FailNext` field in addition to the `Next` field on the non-terminal nodes of the Free.
The Free's `BiBind` and `BiMap` implementations then use this extra field.
It is up to the end-user-built `Interpreter` to make a choice between the two paths based on the outcome of the action.
While implementing it become apparent that it would be nice if the user could decorate which `Free` state represented the shortcutting-fail path so an extra `Fail` attribute was added.

With usage as follows:
```
[Free]
public interface Maybe<A>
{
    [Pure] A Just(A value);
    [Fail] A Nothing();
}
```

```
[Free]
public interface Try<A>
{
    [Pure] A Succ(A value);
    [Fail] A Fail(Exception e);
}
```

```
[Free]
public interface FreeIO<T>
{
    [Pure] T Pure(T value);
    [Fail] T Fail(Error error);
    string ReadAllText(string path);
    Unit WriteAllText(string path, string text);
}
```

If at least one state is decorated with a `Fail` attribute then `BiMap` and `BiBind` will be created for the generated `Free`.

What do you think about this approach in general? The 'FailNext' field? And also the specific 'Fail' attribute?

Apologies for the formatting changes and the removal of some redundant code. I got a bit carried away in the process.

### DIFF
--- a/LanguageExt.CodeGen/FreeAttribute.cs
+++ b/LanguageExt.CodeGen/FreeAttribute.cs
@@ -16,4 +16,10 @@ namespace LanguageExt
     public class FreeAttribute : Attribute
     {
     }
+
+    [AttributeUsage(AttributeTargets.Method, AllowMultiple = false, Inherited = false)]
+    [Conditional("CodeGeneration")]
+    public sealed class FailAttribute : Attribute
+    {
+    }
 }
--- a/LanguageExt.CodeGen/FreeGenerator.cs
+++ b/LanguageExt.CodeGen/FreeGenerator.cs
@@ -722,81 +722,49 @@ namespace LanguageExt.CodeGen
                     });
 
 
-            var freeFuncs = fail != null
-                ? applyToMembers
-                    .Where(m => !m.AttributeLists.SelectMany(a => a.Attributes).Any(IsPureAttr))
-                    .Where(m => !m.AttributeLists.SelectMany(a => a.Attributes).Any(IsFailAttr))
-                    .SelectMany(m =>
-                        new SyntaxNodeOrToken[]
-                        {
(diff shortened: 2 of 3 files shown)


louthy:
Thanks Paul, I’ll get to it as soon as I’m free! :+1:

louthy:
That's a hell of a job, thanks Paul! 👍

Issue #501, opened 2018-10-10
Request: ConfigureAwait(false) within async LanguageExt code (await...)


StefanBertels:
Hi, I had a (obscure) dead-lock problem when using OptionAsync / BindAsync.

The reason was: async code is called by sync code (i.e. Task.Wait or Task.Result or Task.GetAwaiter().GetResult) and there were other things going (probably other async code in a different library) that mattered (consumes threads / task contexts).

I know blocking async code is bad and I should either avoid async calls at all or refactor everything to async.

But the problem can be solved* by calling ConfigureAwait(false) on Task before calling blocking functions (Wait/Result).
I do this in my code (buying the drawbacks) but have still hit a dead-lock when calling BindAsync because BindAsync internally uses awaitwithout ConfigureAwait(false).

Question / suggestion: could LanguageExt use ConfigureAwait(false) so sync callers will not dead-lock?

Background information: it does not dead-lock always, depends on other code around. The critical thing is that without ConfigureAwait(false) the code after await will run on the same context (thread) as the code before. If this context is not available (in use elsewhere) this will block. With ConfigureAwait(false)the code is allowed to continue in another context (thread) which in most cases is no problem (if you don't need the same thread context like in GUI code). This will nearly always result in a working solution (no dead-locks) as long as you don't exhaust the used thread pool -- too many blocking functions (Wait/Result) at the same time.

https://medium.com/bynder-tech/c-why-you-should-use-configureawait-false-in-your-library-code-d7837dce3d7f
https://msdn.microsoft.com/en-us/magazine/jj991977.aspx

bender2k14:
You can create a failing test like I did in #317

bender2k14:
I have put a lot of thought into this. I don't think it is so obvious when to use `ConfigureAwait(false)`. I used to follow the advice from articles like the one to which you linked. In particular, I have had some PRs included into Language Ext that used `ConfigureAwait(false)`. However, I not do sure now.

Suppose you write a failing test and then contribute a PR on which your test passes. Then I think that I can write a test that passes on `master` but fails on your commit. More specifically, my test would include context-sensitive code that I would have executed in the correct context on `master` but in the wrong context on your commit.

As it says in the article to which you linked, my test fail because an exception would be thrown.  In contrast, your test would fail (at least in principle) because a deadlock occurred.

To be clear, I would prefer to debug "my" thrown exception than "your" deadlock. So I do agree with your request to add more calls to `ConfigureAwait(false)`.

[...]

PR #768, opened 2020-04-23
TaskExtensions: Add ConfigureAwait when awaiting any Task

appetere:
**Problem**
I have some synchronous code that has to use the `HttpClient.SendAsync` method.  The code uses the Task monad.  This results in the code deadlocking (almost exactly as described in  https://blog.stephencleary.com/2012/07/dont-block-on-async-code.html).

**Improvement**
By using `ConfigureAwait(false)` throughout the `TaskExtensions` this prevents the deadlock.

**Compatibility**
The change in behaviour is whether the thread resumes on the same context, or not, after an `await`.  So if the code was used _directly_ in an ASP.NET Controller method, or in the UI thread of a Windows Forms application (or similar) which require the context, it could be a problem. But if an intermediate `await` call is made with the default `ConfigureAwait(true)` before using the TaskExtensions then the context would be restored.

### DIFF
--- a/LanguageExt.Core/DataTypes/Task/Task.Extensions.cs
+++ b/LanguageExt.Core/DataTypes/Task/Task.Extensions.cs
@@ -21,7 +21,7 @@ namespace LanguageExt
             if (ma == null) return FailCase<A>.New(Common.Error.Bottom);
             try
             {
-                return SuccCase<A>.New(await ma);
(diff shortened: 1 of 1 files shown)


appetere:
See related issue: https://github.com/louthy/language-ext/issues/501

louthy:
Thanks Steve 👍

PR #774, opened 2020-04-29
Validation serialization was broken

StanJav:
`Validation<string, string>` serialization did not work properly. A string `"2020-04-18T14:28:42.985"` was deserialised into `"04/18/2020 14:28:42"`. 

Note that this works fine with `Either<sting, string>`.

Added unit tests for both Validation & Eiter.


louthy:
Thanks @StanJav ! 👍

PR #779, opened 2020-05-14
add 'flatten' and 'cons' parser helpers

StefanBertels:
I added tests to show potential use cases. They are similar to `sepBy` but `sepBy` does not give me back separator. Maybe there are betters solutions for this. Merge, modify or just close as you like.


louthy:
Thanks @StefanBertels 👍

PR #790, opened 2020-06-13
Stck traverse tests

blakeSaucier:
Some more Traverse tests #728


louthy:
Thanks @blakeSaucier !

PR #794, opened 2020-06-16
Value typed enumerators

louthy:


aalmada:
@louthy Hi! I'm just curious in whether you noticed any significant improvement in performance.

PR #812, opened 2020-10-03
add str<EQ> parser

StefanBertels:
Generic version of a parse I often use to validate e.g. filenames (case insensitive).

This variant parsers char by char and therefore needs an Eq<char> comparer. Therefore I added EqCharOrdinalIgnoreCase, too.


louthy:
Looks good, thanks Stefan!

PR #814, opened 2020-10-03
rename newline parser to LF

StefanBertels:
I stumbled upon "newline" not working like e.g. https://docs.microsoft.com/en-us/dotnet/api/system.io.streamreader.readline which means eating CRLF and LF.

Seems the definition in LanguageExt is directly based on Haskell and I guess you see a value in this.
But I think some of those definitions are uncommon / unexpected from a .NET perspective. This might be a reason to change or rename this.

The commit here should avoid confusion for `newline` (`LF` is quite explicit).

I personally think `endOfLine` (which is what I should have used in the first place) isn't the best name for LF/CRLF. I personally would expect a parser equalivalent to regex `$` (which is `eof` in parsec). Maybe those should be renamed to more clear names, too?

Another part I find misleading is `space`. This should be named `whitespace`. Of course I could use `ch(' ')` but I expected `space` to be 0x20 only and there is e.g. `tab`...

But I'm ok with closing this if you think it really should be like before (at least consistent for Haskell) and of course changes might break user code.

### DIFF
--- a/LanguageExt.Parsec/Parsers/Char.cs
+++ b/LanguageExt.Parsec/Parsers/Char.cs
@@ -20,13 +20,13 @@ namespace LanguageExt.Parsec
             spaces = skipMany(space).label("white space");
             control = satisfy(System.Char.IsControl).label("control");
             tab = satisfy(c => c == '\t').label("tab");
-            newline = satisfy(c => c == '\n').label("lf new-line");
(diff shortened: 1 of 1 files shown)


louthy:
This is a breaking change and will likely affect most users of Parsec, so I'm going to push back on this.  I wouldn't be against either supporting both or making `newline` `[Obsolete]` with an advice message.

PR #820, opened 2020-10-25
adds `LF` parser and marks `newline` as obsolete

StefanBertels:
See https://github.com/louthy/language-ext/pull/814.

This adds an explicit `LF` parser to be used instead of `newline` (which might be selected by mistake by .NET developers, especially when domesticated on Windows).


louthy:
Thanks @StefanBertels 👍

PR #822, opened 2020-10-27
make all UnitsOfMeasure types implement IComparable

StefanBertels:
This is a suggestion.

Currently those types only implement `IComparable<T>` which is a good thing in theory because of type-safety.
But there is still code out there that uses `IComparable`.

Example: [`Assert.InRange(result, min, max)`](https://github.com/xunit/assert.xunit/blob/main/RangeAsserts.cs#L26-L30) in xunit.

I started a [discussion there](https://github.com/xunit/xunit/discussions/2176), to maybe add support for `IComparable<T>`-only types but I'm not sure what's the best way to solve this more generally. At least dotnet core itself still implements the untyped variant for e.g. [`int32`](https://github.com/dotnet/runtime/blob/master/src/libraries/System.Private.CoreLib/src/System/Int32.cs).

In LanguageExt there are already a lot of types (like `Option`) that implement `IComparable`...

### DIFF
--- a/LanguageExt.Core/DataTypes/UnitsOfMeasure/Accel.cs
+++ b/LanguageExt.Core/DataTypes/UnitsOfMeasure/Accel.cs
@@ -14,7 +14,8 @@ namespace LanguageExt.UnitsOfMeasure
     /// </summary>
     public struct Accel :
         IComparable<Accel>,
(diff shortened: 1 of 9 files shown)


louthy:
Thanks @StefanBertels 👍

PR #825, opened 2020-11-03
Add missing Serializable attribute on abstract base class in union codegen

nekronos:
Serializing a union where an abstract class is used throws exception as the generated intermediate base class is not marked as `Serializable`.  This PR makes the codegen add the missing attribute


louthy:
Looks good, thanks!

PR #827, opened 2020-11-05
add fluent matchers to OptionAsync

### LINKED ISSUE #821: OptionAsync missing .Some and .None fluent matchers
There are .Some and .None fluent matchers present on Option, TryOption, and even TryOptionAsync, but not on OptionAsync.  Maybe there's a good reason, but it seems like an oversight.

eldon111:
Resolves #821


louthy:
Looks good, thanks 👍

PR #834, opened 2020-11-25
Should either(str, str) be nonconsuming?

StefanBertels:
I stumbled upon a parser like this that failed (unexpectedly):

```
var twoThree= either(str("two"), str("three"));
Assert.Equal("three", parse(twoThree, "three").ToOption()); // currently fails
```

This is caused by `str` consuming "t" of "two" before trying to parse "three" (on "hree", which fails).

Technically everything is fine, I read the comment in Prim.cs and I use `attempt` to get around this. 

One "solution" could be that (by convention) a parser that has its own label should not consume parts.

### DIFF
--- a/LanguageExt.Parsec/Parsers/Char.cs
+++ b/LanguageExt.Parsec/Parsers/Char.cs
@@ -251,13 +251,13 @@ namespace LanguageExt.Parsec
         /// Parse a string
         /// </summary>
         public static Parser<string> str(string s) =>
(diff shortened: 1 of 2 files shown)


louthy:
the existing behaviour is expected, you've consumed something, which is always a complete failure.  And so you need to use `attempt`, which usually should be used for anything where there's choice, or many items parsed.  It's arguable that `attempt` should be implied, but that's how the original Parsec library in Haskell works, so I decided to not change it.
```c#
    either(attempt(str("two")), str("three"));
```

PR #835, opened 2020-11-25
add missing struct constraint

StefanBertels:
add missing struct constraint


louthy:
Thanks @StefanBertels !

PR #841, opened 2020-12-11
Eq.ToEqualityComparer (Ord.ToComparer)

StefanBertels:
This is an addition to #720. I think it's less important than `IComparer` but still has some use when dealing with non-LanguageExt code (like LINQ extensions, System.Interactive ...).

I've seen that you changed #720 to avoid a record problem and I added a small assertion in that ord test to make that more explicit (avoiding that it get's changed back easily).

I renamed `Ord.ToComparable` to `Ord.ToComparer` because I think the latter is more correct to align this to *Comparable vs. *Comparer and *Equatable vs. *EqualityComparer.


StefanBertels:
I personally don't need https://github.com/louthy/language-ext/pull/841/commits/3c131cc821e648c607325e8ef4880a17325f75dd but added this just in case we want avoid any breaking change.

StefanBertels:
rebased on current main

louthy:
Thanks @StefanBertels 👍

PR #842, opened 2020-12-11
Not a bug, just two tests (was: `Map` with target net5.0 and unicode)

StefanBertels:
I think there is a bug somewhere in `Map` for net5.0.
Everything is fine for netcoreapp3.1.

Note: I found this using property based testing and can recommend this (especially for generic code like LanguageExt). Second commit is request to add `FsCheck.Xunit` to enable this kind of tests.


StefanBertels:
Seems that \u0000 fails on all targets.

StefanBertels:
Found the cause:

Default comparer for strings is quite different from "Ordinal" if unicode characters are used (starting with net5.0).
https://docs.microsoft.com/en-us/dotnet/standard/base-types/string-comparison-net-5-plus

I fixed the tests and everything looks good. Maybe you want to add these tests (otherwise just close).

louthy:
Not sure how this is a lang-ext issue?  Isn't this just about picking the comparer you want?

Issue #844, opened 2020-12-15
New Library - Wraps repetitive code with LanguageExt


gregberns:
Hi Paul, 
Recently I've joined a company as tech lead and am rebuilding a platform with C#. Obviously LanguageExt is playing a centrall roll in the code base. 

After the 5th small service I found I was re-writing/copy-pasting a large portion of the code like: wrapping `GetEnvironmentalVariable` with Validation, wrapping HttpClient, JSON and Yaml deserialization, and a paging in REST APIs.

So I took all the common code and aggregated it into: [LanguageExtCommon](https://github.com/gregberns/languageextcommon). We've now used it for the last couple weeks and cut out a ton of repetitive code in our services.

Its still very much in alpha, but I'd image we'll add things like database wrappers (maybe as separate nuget's), Azure SDK, who knows... 

Anyways, I just wanted to share this and get any feedback, suggestions, or ideas you may have. Hell, if there's a better name for the project I'm open to it!

https://github.com/gregberns/LanguageExtCommon

cc @StefanBertels (since you seem very active the last couple years)

StefanBertels:
Hi Greg,

IMHO: I think a single helper library is problematic because it might result in "endless scope" and incompleteness.

If there is a general feature which seems missing in `core` (or `parsec`) this should be discussed/added via a PR to Pauls core library. Of course some features might be too special / opinionated / adding unwanted dependencies or not general enough. E.g. I would like to have functions in LanguageExt.Core for getting e.g. `Either<Error, T>` from any `T` returning function without the "need" to use custom or third party code to build an expression like `try/catch => Try<T> => Either<Exception,T> => Either<Error,T>`. If the built-in way is too much boilerplate, let's make it easier. But maybe it's not boiler plate but just clarity/explicitness.

If you solved a specific thing like e.g. interoperability with Azure SDK I suggest to create a helper package specifically for that use case and make them available independently of each other. (Probably you should check with Paul about naming it `LanguageExt*`.) I created some similar classlibs (internally, non-public) adding e.g. support for `Option` types and parsers in `CsvHelper` or get json reader/writer support for `Json.NET` and `System.Text.Json` (supporting more familar json formats). Doing this in a public library IMHO should aim to build a thing that isn't "incomplete" and should try to avoid being opinionated and avoid adding new types (like `HttpResponse`).

[...]

PR #849, opened 2021-01-14
Add EnvironmentIO effects

gregberns:
Add Eff support for `System.Environment`. ~~There's still some cleanup on the comments that needs to be done.~~

It was [also mentioned](https://github.com/louthy/language-ext/issues/844#issuecomment-759482845) not to return `Option`. I assume we could use `getEnvironmentVariable` as an example. In that case, it seems a little weird because it not existing isn't necessarily an error. But I'm still not clear on how to handle/deal with success, expected errors, and exceptional errors.


louthy:
Thanks Greg.  I've slightly changed the Aff/Eff stuff, so I'll accept this and refactor it to the new setup.

PR #872, opened 2021-04-29
Change NumTypeExtensions.Sum to use Reduce instead of Fold

### LINKED ISSUE #869: NumTypeExtensions.Sum might be incompatible with the PRED of the type being summed
Consider the following `NumType` derivate:

```
public class PositiveNumber : NumType<PositiveNumber, TInt, int, GreaterThan<TInt, int, I0>>
{
    public PositiveNumber(int value)
        : base(value)
    {
    }
}
```

When trying to `.Sum()` a list of these, you will get an exception because the implementation uses `Fold` and tries to use a seed value of `0` with the same `PRED` as the type being summed.

I could work around it using:

```
public static class PositiveNumberExtensions
{
    public static PositiveNumber Sum(this IEnumerable<PositiveNumber> xs)
        => self.Aggregate((x1, x2) => LanguageExt.Prelude.add(x1, x2));
}
```

Easy fix, but it would be nice if LanguageExt actually did this right out of the box (cause it bites you at run-time).

mario-d-s:
Fixes #869; depending on the predicate of the NumType the original implementation could throw an exception (the seed value of "0" might violate the predicate condition).

@TysonMN turns out that other "Fold" overload already exists and is called "Reduce" :)

I was thinking about adding tests but I'm not sure if it's desirable for such a simple function (I found no original ones for this), let me know if you want any.

### DIFF
--- a/LanguageExt.Core/DataTypes/NumType/NumType.Extensions.cs
+++ b/LanguageExt.Core/DataTypes/NumType/NumType.Extensions.cs
@@ -8,12 +8,12 @@ namespace LanguageExt
         public static SELF Sum<SELF, NUM, A>(this IEnumerable<NumType<SELF, NUM, A>> self)
             where SELF : NumType<SELF, NUM, A>
             where NUM : struct, Num<A> =>
-            self.Fold(NumType<SELF, NUM, A>.FromInteger(0), (s, x) => s + x);
+            (SELF) self.Reduce((s, x) => s + x);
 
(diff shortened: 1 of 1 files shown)


louthy:
This fails with an empty enumerable, and so it doesn't maintain the existing behaviour and would be a breaking change.

I'd propose a second `Sum` method that uses the reduce method, perhaps called `NonEmptySum`, or something that clearly denotes the requirement of the `IEnumerable` argument.  It should return an `Option<SELF>` also.

louthy:
Closing for now.  Feel free to re-open if you want to deal with issues raised.

Issue #831, opened 2020-11-19
OptionAsync impicit cast throws error when value is null


Lonli-Lokli:
I am getting exception with 3.5.20-beta while trying to use OptionAsync<T> as a wrapper over nullable result
```

public class Test
    {
        public string value => "test";
    }
    class Program
    {
        static async Task Main(string[] args)
        {
            await test();
            Console.ReadKey();
        }

        private static async Task test()
        {
            Console.WriteLine("Test started");
            Console.WriteLine("Wrap for some: {0}", (await wrap(true))?.value ?? "ERROR");
            Console.WriteLine("Wrap for none: {0}", (await wrap(false))?.value ?? "Empty");  // throws
            
            Console.WriteLine("Test ended"); 
        }
        private static OptionAsync<Test> wrap(bool some)
        {
            return some ? delaySome() : delayNone();
        }
        
        private static async Task<Test> delaySome()
        {
            await Task.Delay(2000);
            return new Test();
        }
        
        private static async Task<Test> delayNone()
        {
            await Task.Delay(2000);
            return null;
        }

```

BUG:
Test started
Wrap for some: test
Unhandled exception. LanguageExt.ValueIsNoneException: Value is none.
   at LanguageExt.OptionAsyncAwaiter`1.GetResult()
   at LanguageExt_OptionAsync.Program.test() in D:\Projects\LanguageExt_OptionAsync\LanguageExt_OptionAsync\Program.cs:line 23
   at LanguageExt_OptionAsync.Program.Main(String[] args) in D:\Projects\LanguageExt_OptionAsync\LanguageExt_OptionAsync\Program.cs:line 15
   at LanguageExt_OptionAsync.Program.<Main>(String[] args)

TysonMN:
This is because the implicit converter for `Task<A>`....
https://github.com/louthy/language-ext/blob/3799d385a88ab100af7f64e21eb154101db4c338/LanguageExt.Core/DataTypes/OptionAsync/OptionAsync.cs#L148-L149
...calls this constructor that expect the value in the `Task<>` to be not be `null`.
https://github.com/louthy/language-ext/blob/3799d385a88ab100af7f64e21eb154101db4c338/LanguageExt.Core/DataTypes/OptionAsync/OptionAsync.cs#L85-L86

To match the behavior for `Option<>`, I do think an instance of `OptionAsync<>` should be returned in the `None` state.  Do you agree @louthy?

P.S.
[...]

PR #873, opened 2021-05-17
Do not throw exception on implicit Task<NULL> conversion

Lonli-Lokli:
Fixes https://github.com/louthy/language-ext/issues/831


louthy:
Thanks @Lonli-Lokli !

PR #957, opened 2021-11-23
fix: AtomSeq<A> now is IEnumerable<A>

StefanBertels:
AtomSeq didn't implement generic IEnumerable<A>, test added.


louthy:
Good spot, thanks @StefanBertels  :)

Issue #923, opened 2021-10-12
Request: `AtomChangedEvent<A> Change` for `AtomHashMap<K, V>` and `Ref<T>`


CK-LinoPro:
`Atom<T>` has an event `AtomChangedEvent<T> Atom<T>.Change` that emits any changes made to the atom via `Atom<T>.Swap(...)`.
Is it possible to provide the same event for `AtomHashMap<K, V>` and `Ref<T>`?

CK-LinoPro:
Do you have an update on this topic?
Currently, what I'm doing is a error-prone workaround where I update an `Atom<T>` each time I update its `Ref<T>` instance,
but at that time the value inside the `Ref<T>` instance has not been written yet.

If you like I could look into it myself and provide a pull request.

CK-LinoPro:
I've created a pull request: https://github.com/louthy/language-ext/pull/972.

CK-LinoPro:
Thanks for your review of the PR, it's unfortunate my approach is too naive to be accepted.
I could convince my team to work on a fork for now, so there is no hurry for a more practical solution.
I know this request isn't without issue as the whole topic of STM is rather complicated.

louthy:
[I've made a start with this](https://github.com/louthy/language-ext/commit/9552ef54e2bcb4403b83f57978daaa7384213b57).  I am slightly short of time right now, so I've just done the foundational work and implemented change-tracking for `AtomHashMap<K, V>`.

The underlying `TrieMap<EqK, K, V>` now has variants of all collection-transforming methods that can track changes, i.e. `Add` and `AddWithLog`.  The `*WithLog` variants will tracking insertions, deletions, and updates - even for collection wide transformations.

The `AtomHashMap<K, V>` has been updated to use these methods if the `OnChange` event is not-null.  That means for those who aren't using change-tracking feature there's no overhead.  Even for those that are, it's very optimal, and does the minimum work needed to collect the changes.

The event type looks like this:
```c#
    public delegate void AtomHashMapChangeEvent<K, V>(
        HashMap<K, V> Previous, 
        HashMap<K, V> Current,
        HashMap<K, Change<V>> Changes); 
```
So, for each atomic change you get the previous state of the hash-map, the current-state, and a `HashMap` of changes - which can be:

* `ItemUpdated<V>` - which has the old-value and the new-value for an updated key,
[...]

PR #972, opened 2022-01-04
Add Changed events for Ref, AtomHashMap and AtomSeq

CK-LinoPro:
`Atom<A>` has a `Changed` event that is very useful for reactive code.
I've added the same kind of event for `Ref<A>`, `AtomHashMap<TEq, K, V>`, `AtomHashMap<K, V>` and `AtomSeq<A>`.

This PR was previously discussed in https://github.com/louthy/language-ext/issues/923.

### DIFF
--- a/LanguageExt.Core/Concurrency/AtomHashMap/AtomHashMap.Eq.cs
+++ b/LanguageExt.Core/Concurrency/AtomHashMap/AtomHashMap.Eq.cs
@@ -29,6 +29,8 @@ namespace LanguageExt
     {
         internal volatile TrieMap<EqK, K, V> Items;
 
+        public event AtomChangedEvent<HashMap<EqK, K, V>>? Change;
+
         /// <summary>
         /// Creates a new atom-hashmap
         /// </summary>
@@ -116,6 +118,7 @@ namespace LanguageExt
                 }
                 if (ReferenceEquals(Interlocked.CompareExchange(ref Items, nitems, oitems), oitems))
                 {
+                    this.Change?.Invoke(new HashMap<EqK, K, V>(nitems));
                     return default;
                 }
                 else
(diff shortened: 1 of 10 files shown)


louthy:
This adds a big performance overhead to the current system, which is very much designed to be as fast as possible (whilst maintaining the atomic integrity, of course).

There's a few issues I see here:

* How useful really is it that `AtomHashMap` and `AtomSeq` pass a `HashMap` out of the event?  It isn't a hash-map of changes, it's just the new map.  It's no different to firing an event with a value of `Unit`, as all it indicates is that "something has changed", not what has changed.

* `Ref` subscribes to the changes in the 'world' state of the STM:
  * The static constructor for `STM` calls `Map` on the entire state of the world on every transaction.  That's a loop of `n` for every transaction, where `n` is the number of `Ref`s in existence.
  * `OnSTMChange` will do up to four allocations per-mutation, due to the lambdas.
  * So that's `4 * n` possible allocations per transaction.

Unfortunately, I think this approach is too naïve, and will lead to a less stable system overall, because:
   * The transactions will take much longer to run, so there's likely to be more conflicts, and therefore re-committing of the transaction
   * The linear scaling of allocations based on the number of refs will really hurt the GC for serious use-cases

My early thoughts with this was to create a transaction-log ADT for each operation performed.  For example: 

```Haskell
    AtomicMapOperation k v
        = AddItem k v
        | UpdateItem k v
        | RemoveItem k

    AtomicSetOperation k
        = AddItem k
        | UpdateItem k
        | RemoveItem k

    AtomicSeqOperation a
        = AddItem int a
        = RemoveItem int
```
Then any transaction could collect them and push them out to a event-stream when successful.  It's not trivial to do this though, and requires some effort in the `TrieMap` and `TrieSet` types (which is why I was putting it off until I had a clear idea of how to do this without causing performance headaches).

This doesn't help you, I know.  One thing I'd probably advise is to write your own wrappers for now that wrap these types, and expose the events you need, because I can't sign this off in its current form.

PR #974, opened 2022-01-17
Clarify scope of compatibility qualification.

ode2code95:
Auto-properties are also field-backed properties.


louthy:
Thanks @ode2code95 👍

Issue #991, opened 2022-02-11
Unexpected Func<> argument order difference in Map vs. System.Linq.Select


aage:
Using the `Map` overload with `index` within `LanguageExt` the other day, I introduced an unexpected bug in my code whilst mapping an `IEnumerable<int>`; coming from `System.Linq` I expected that the `index` would be the _second_ argument in the `Func<>` that needs to be passed.

In `System.Linq.Select` this is the `selector`: `Func<TSource, int, TResult> selector` (as seen in [this file](https://github.com/dotnet/corefx/blob/master/src/System.Linq/src/System/Linq/Select.cs) on line 65). Whilst in `LanguageExt` 'tis the other way around.

I think this difference would be clearer if it would be featured in the comments on relevant functions within `LanguageExt` so a developer checking out Intellisense would be informed.

I would be happy to submit a PR which add this nuance to the comments of these functions, if you agree this would be an improvement.

louthy:
> I would be happy to submit a PR which add this nuance to the comments of these functions, if you agree this would be an improvement.

Documentation and PRs are always appreciated 👍

aage:
Added a PR:
https://github.com/louthy/language-ext/pull/1001

louthy:
This is now released

PR #1001, opened 2022-03-05
Clarify difference between 'Map' function and 'Select' within Linq, concerning the index argument #991

aage:


louthy:
Thanks @aage 👍

Issue #999, opened 2022-02-23
Velocity.KilometresPerHour is incorrect


YuriiNskyi:
```cs
public static Velocity KilometresPerHour(this double self) =>
    new Velocity(self * 1000.0 * 3600.0);
```

I'm not sure that this formula is correct.

Internally value stored as metres per second, so `1.KilometresPerHour()` must be `0.27`, not current value `3600000`.

The working formula is: `(self * 1000) / 3600`

Actually, `MilesPerHours` method has the same typo: we must divide by 3600, rather than multiplying by it.

YuriiNskyi:
@louthy sure, please, check this https://github.com/louthy/language-ext/pull/1005

louthy:
This is now released in `v4.0.7-beta`.  Thanks for the PR :)

PR #1005, opened 2022-03-09
Fix incorrect creation of Velocity

YuriiNskyi:
Fix issue https://github.com/louthy/language-ext/issues/999


louthy:
Thanks @YuriiNskyi 👍

Issue #997, opened 2022-02-21
Suggestion: Use Array.Empty to initialize empty arrays


JonathanQuinth:
Would replacing new T[0] with Array.Empty<T>() be an easy optimization to prevent unnecessary allocations?

See for instance EnumerableOptimal.cs where empty arrays are created in the BindFast functions.

louthy:
I'd be very surprised if `new T[0]` actually allocates anything.  Have you checked the generated IL?

JonathanQuinth:
Yes, I think it does. At least it generates the "newarr" instruction which is supposed to create a new array (albeit zero-length).

The IL instructions I get for EnumerableOptimal.BindFast using JetBrains dotPeek:
IL_0000: ldarg.0  
IL_0001: brfalse.s
IL_0003: ldarg.0  
IL_0004: ldarg.1  
IL_0005: newobj   
IL_000a: stloc.0  
IL_000b: ldloc.0  
IL_000c: ret
IL_000d: ldc.i4.0
IL_000e: newarr <--
IL_0013: stloc.0  
IL_0014: ldloc.0 
IL_0015: ret

Array.Empty<T>() merely returns a reference to a static zero-length array and does not allocate anything. I have not done any benchmarks myself, but here is one I found through Google: [https://www.tabsoverspaces.com](https://www.tabsoverspaces.com/233833-best-way-to-create-an-empty-collection-array-and-list-in-csharp-net)

louthy:
Huh, I always thought the C# compiler team would be smarter than that.  That'll teach me to assume!

Anyway, that looks like a fair optimisation if you want to submit a PR

MrWuffels:
> Huh, I always thought the C# compiler team would be smarter than that. That'll teach me to assume!
> 
> Anyway, that looks like a fair optimisation if you want to submit a PR

To be fair to the C# compiler team, this is on purpose. 
While `object.ReferenceEquals(Array.Empty(), Array.Empty())` returns true, `object.ReferenceEquals(new T[0], new T[0])` does not.

PR #1010, opened 2022-03-15
Remove allocation overhead for empty arrays (and enumerables) (#997)

timmi-on-rails:
This fixes issue #997 and a few other places that I considered performance critical.
Would be interesting to actually benchmark this.


louthy:
Looks good @timmi-on-rails , Thanks! 👍 

I'll get this in the next release.

PR #1026, opened 2022-04-26
Bugfixes: SeqLazy.Equals sometimes wrong, Set.Case returned wrong data/type

StefanBertels:
Two bug fixes:
* `Set.Case` returned wrong type (not `var x or var (x,xs)`)
* `SeqLazy` had broken `GetHashCode()` which could result in wrong Equality (two `Seq` could be equal with `Equals` returning false).


StefanBertels:
Note: Maybe property based testing would be a good fit because it can help find edge cases (like this one). Maybe add FsCheck.Xunit to allow this kind of tests?

louthy:
Thanks @StefanBertels !

PR #1045, opened 2022-06-04
[Feature] - Enhanced Schedule

bmazzarol:
This enhances the existing Schedule bringing its features closer to that provided by ZIO and Polly.

A schedule is defined as a potentially infinite stream of positive durations. This is simpler than the ZIO design (IMHO) and provides for the vast majority of features it provides. I also think its easy to understand and extend.

The other additional element is a schedule transformer, this is just a Func<Schedule,Schedule>. This provides for transformations that potentially require the construction of a whole new Schedule from the current one.

This does not provide a means to create stateful schedules that consume the effect result and environment (ZIO can do this I think). 
I haven't got a sold mental model for that sort of schedule at this stage (IEnumerable does not fit), but do have a requirement for one.

The usage should be the same and (I hope) the provided schedules are correct.

@louthy I understand that no issue was raised as per the contribution guidelines, but I have derived so much value from this library that I wanted to give something back.

Have raised as a draft and feedback is welcome as well as not proceeding with any of it.


louthy (changes requested):
Firstly, thanks so much for this effort, it's really, really appreciated!  When I was writing the effect system I always intended to build this fully featured schedule system, but had 1000s of things to build (in the effects subsystem), so just created something quickly that could be replaced later.  This is exactly what I would have looked to build, and it looks great.

There's a few style comments and some questions about naming in the review notes.  A few more things:

* Can you add `#nullable enable` to the top of each file?  I'm adding nullable-reference support a file at a time at the moment.  It's not ideal, but the big bang approach doesn't work for this lib, so any new or refactored code should have that at the top.
* Drop the use of `private`, it's never needed and is just a bit of extra clutter in C# land
* Always have a line-break after the `=>` on expression bodied members.  It helps keep the declaration line clear and easy to read.
* Are there any breaking changes or changes of behviour for those using the current `Schedule` implementation?

Again, great work.  I'm looking forward to getting this into the lib.

LanguageExt.Core/Effects/Schedule/Schedule.Contructors.cs
```diff
@@ -0,0 +1,574 @@
...
+    /// </remarks>
+    /// <param name="medianFirstRetryDelay">median delay to target before the first retry, call it f (= f * 2^0).
+    /// Choose this value both to approximate the first delay, and to scale the remainder of the series.
+    /// Subsequent retries will (over a large sample size) have a median approximating retries at time f * 2^1, f * 2^2 ... f * 2^t etc for try t.
+    /// The actual amount of delay-before-retry for try t may be distributed between 0 and f * (2^(t+1) - 2^(t-1)) for t >= 2;
+    /// or between 0 and f * 2^(t+1), for t is 0 or 1, default 100 milliseconds</param>
+    /// <param name="retry">The maximum number of retries to use, default 5</param>
+    /// <param name="seed">An optional seed to use</param>
+    /// <param name="fastFirst">flag to indicate the first retry is immediate, default false</param>
+    public static Schedule PollyDecorrelated(
```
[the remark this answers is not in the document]
bmazzarol: Removed all Polly releated anything.

I have create my own poor mans decorrelation transformer instead.

bmazzarol:
@louthy I want to add some tests to cover the Run methods against the effects. Will try get that done tonight.

bmazzarol:
>     * Are there any breaking changes or changes of behviour for those using the current `Schedule` implementation?
> 

Unfortuntly there is (I think), the fibonnaci differs. Please have a look through my tests here and make sure that I have done it correctly LanguageExt.Tests/ScheduleTest/ScheduleChangesTests.cs

LanguageExt.Core/LanguageExt.Core.csproj
```diff
@@ -5,10 +5,18 @@
   <PropertyGroup Label="Configuration">
```
bmazzarol, line 5: @louthy I will fix this not sure what happend with my merge

bmazzarol:
> > ```
> > * Are there any breaking changes or changes of behviour for those using the current `Schedule` implementation?
> > ```
> 
> Unfortuntly there is (I think), the fibonnaci differs. Please have a look through my tests here and make sure that I have done it correctly LanguageExt.Tests/ScheduleTest/ScheduleChangesTests.cs

No was my bad, will update the tests

LanguageExt.Core/LanguageExt.Core.csproj
```diff
@@ -56,4 +70,87 @@
...
+    </None>
+    <None Update="HKT\AsyncSync\HKT.AsyncSync.Extensions.tt">
+      <Generator>TextTemplatingFileGenerator</Generator>
+      <LastGenOutput>HKT.AsyncSync.Extensions.cs</LastGenOutput>
+    </None>
+  </ItemGroup>
+  <ItemGroup>
+    <Service Include="{508349b6-6b84-4df5-91f0-309beebad82d}" />
+  </ItemGroup>
 </Project>
```
louthy, line 59: You'll need to revert all changes to this file, this shouldn't need to be updated.  Presumably it was a merge issue of some sort?

louthy (changes requested):
Getting close!  A few minors, and one major (the changes to LanguageExt.Core.csproj need reverting).

LanguageExt.Core/Effects/Schedule/Schedule.Contructors.cs
```diff
@@ -0,0 +1,446 @@
...
+    [Pure]
+    private static int DurationToIntervalStart(int intervalStart, int currentIntervalPosition, int intervalWidth)
+    {
+        var steps = intervalStart - currentIntervalPosition;
+        return steps > 0 ? steps : steps + intervalWidth;
+    }
+
+    [Pure]
+    private static int RoundBetween(int value, int min, int max) =>
+        value > max ? max : value < min ? min : value;
```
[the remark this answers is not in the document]
bmazzarol: Agreed Math.Clamp is what I wanted here. Looks like its added in 6? https://docs.microsoft.com/en-us/dotnet/api/system.math.clamp?view=net-6.0

So formatted as requested.

bmazzarol:
@louthy there is one area where the results for the schedules differ (that I found), Intersect looks to behave differently. Let me know your thoughts.

Happy to add some more test cases there to explore more potential differences.

louthy:
> Intersect looks to behave differently. Let me know your thoughts.

I'm less bothered about that I reckon, but if you can document the differences I can highlight it when I do the release notes.  It's important that I explain what's going to change for the end-users.

LanguageExt.Core/Effects/Schedule/Schedule.Contructors.cs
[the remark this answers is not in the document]
louthy: That's weird I just kinda assumed that `Math.Clamp` had always been there!  Let's go with the better formatted ternaries then

bmazzarol:
@louthy completed a final review I'm happy.

In regards to the change in behaviour for intersect I would describe it like this,

With the existing code, the intersection of schedules, altered the seed duration of the schedule with the lower starting seed duration. That schedule then continued on with the sequence based on the new higher seed duration from the other schedule.

Schedules are now isolated, they run together and the intersection/union of them, produces a new schedule that is feed by, but does not influence, the two schedules.

LanguageExt.Core/Effects/Schedule/Schedule.cs
```diff
@@ -13,9 +13,9 @@ namespace LanguageExt;
 /// of schedulers and then union them using `|` or intersect them using `&amp;`.  Union will take the minimum of the two
 /// schedules to the length of the longest, intersect will take the maximum of the two schedules to the length of the shortest.
 ///
-/// Any IEnumerable&lt;Duration&gt; can be converted to a Schedule using ToSchedule() or Schedule.FromDurations(...).
-/// Schedule also implements IEnumerable&lt;Duration&gt; so can make use of any transformation on IEnumerable.
-/// A Schedule is a struct so an AsEnumerable() method is also provided to avoid boxing.
+/// Any `IEnumerable&lt;Duration&gt;` can be converted to a Schedule using `ToSchedule()` or `Schedule.FromDurations(...)`.
+/// Schedule also implements `IEnumerable&lt;Duration&gt;` so can make use of any transformation on IEnumerable.
+/// A Schedule is a struct so an `AsEnumerable()` method is also provided to avoid boxing.
```
louthy, line 18: You can use the `<` and `>` characters instead of `&lt;` and `&gt;`.  The [API doc-gen tool](https://github.com/louthy/best-form) is smart enough to know what are tags and what's pure documentation, and deals with them correctly.  Never been a fan of putting these special XML characters in to documentation read by programmers, it completely blinds the reader.

https://louthy.github.io/language-ext/LanguageExt.Core/index.html

louthy (changes requested):
Small change to the documentation here

louthy:
> @louthy completed a final review I'm happy.
> 
> In regards to the change in behaviour for intersect I would describe it like this,
> 
> With the existing code, the intersection of schedules, altered the seed duration of the schedule with the lower starting seed duration. That schedule then continued on with the sequence based on the new higher seed duration from the other schedule.
> 
> Schedules are now isolated, they run together and the intersection/union of them, produces a new schedule that is fed by, but does not influence, the two schedules.

Thanks for that 👍   I'll try and show some examples when I do the release.

LanguageExt.Core/Effects/Schedule/Schedule.cs
bmazzarol: Done

louthy:
All looking good. Thanks for the effort here and thank you for taking the review notes with humility, it's a substantial chunk of work which is vary valuable 👍

PR #1051, opened 2022-06-10
Minor Documentation/Access level Fixes for Schedule

bmazzarol:
* Fixed incorrect/inconsistent documentation 
* Added Filter/Where
* Made Bind/SelectMany and Map/Select accessible


louthy:
Good stuff, thanks Ben 👍

PR #1063, opened 2022-06-22
Fixes #1062: Added .Bind() function

### LINKED ISSUE #1062: Mapping an operation which itself returns Result<T> returns a Result<Result<T>>
Consider the following code (I have not compiled it so it may be wrong, but should illustrate the point):

```
public interface ISomething<T>
{
    Result<T> DoSomething(int x);
}

public class Runner<T>
{
    private ISomething<T> _something; // Assume this is injected through a constructor

    // Constructor removed for brevity

    public void Run()
    {
        var foo = new Result<int>(123);
        var bar = foo.Map(x => _something.DoSomething(x));
    }
}
```

The desired return value for 'bar' would ideally be the `Result<T>` from the mapped operation. Instead it is of type `Result<Result<T>>`.

Scott Wlaschin (swlaschin) demonstrates a _Bind_ operation in his F# presentations which gives the desired behaviour.

dale-palmer:
When trying to run `.Map()` for something that already returns a `Result<T>` the result is of type `Result<Result<T>>`.

I have therefore added a `.Bind()` function which will return the `Result<T>` itself without being wrapped into another `Result<T>` type. This resolves the issue I raised in #1062

### DIFF
--- a/LanguageExt.Core/Common/Result/Result.cs
+++ b/LanguageExt.Core/Common/Result/Result.cs
@@ -184,6 +184,18 @@ namespace LanguageExt.Common
                 : new Result<B>(await f(Value));
 
         [Pure]
+        public Result<B> Bind<B>(Func<A, Result<B>> f) =>
+           IsFaulted
+               ? new Result<B>(Exception)
+               : f(Value);
+
+        [Pure]
+        public async Task<Result<B>> BindAsync<B>(Func<A, Task<Result<B>>> f) =>
+            IsFaulted
+                ? new Result<B>(Exception)
+                : (await f(Value));
+
+        [Pure]
         public int CompareTo(Result<A> other) =>
             default(OrdResult<A>).Compare(this, other);
 
--- a/LanguageExt.Tests/ResultTests.cs
+++ b/LanguageExt.Tests/ResultTests.cs
@@ -64,7 +64,7 @@ namespace LanguageExt.Tests
                 Succ: v => v,
                 Fail: ex => -1
             );
-            
+
             Assert.Equal(1, output);
         }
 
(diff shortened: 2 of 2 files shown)


dale-palmer:
@louthy I am bumping this as it seems to be going stale. Please could you take a look and give me feedback as necessary?

Many thanks.

louthy:
I’m in two minds.  If you read the comment at the top of the Result class you’ll see Result was never meant to be consumed.  But I have no choice due to Try etc being a delegate.  Which is why I built the other function based monads as structs that wrapped a delegate.

This request comes up so often that I need to make one of two choices:

* Remove the public interface for Result altogether (and direct everyone to Fin, which is Result on steroids)
* Make Result a first class citizen on Lang-ext

The first option is cheap for me, with no ongoing maintenance costs.  But probably some annoyed users.  The second is a permanent drain on my time did little gain other than user happiness.

So, if I’m slow to respond to any PR, it’s usually because I want to sit on it first a while and decide what direction I want to take.  

And to be clear, these changes are not enough to provide parity with the other types in the lib.  So it would need more effort.

dale-palmer:
@louthy : Thank you for getting back to me. That makes sense. I'll leave it to you to decide what to do with the proposal.

For what it's worth, I've written this as an extension method a few times now which is why I wanted to contribute it but it seems that I have just been using your library wrong all this time.

One thing I will say is that I've seen it presented a few times by some YouTubers who have large followings so removing the Result type may be problematic, but change always is, right? That being said, I did always find it odd that it wasn't in the main namespace like Option is. I guess this explains why.

louthy:
Closing this for now.  The recommendation is to use `Fin` instead of `Result`.

PR #1078, opened 2022-06-30
Fix Distinct with custom comparer (hash bug)

StefanBertels:
There is a bug in .Distinct() when a custom comparer is used. PR includes test and fix.

Bug is that default GetHashCode is used -- this is not valid with a custom (key) comparer.

First both commits only fix one occurence of the bug (there are more). Some things come to my mind (before adding more tests/fixes):

1. IMHO there should be a `Distinct<EQ, T, K>(...)` variant which I would prefer for my use case (avoiding runtime comparer).
2. Maybe avoid 'func parameter' style `Func<string,string,bool>` for helper functions like this (comparer). This would avoid hiding a gethashcode "problem" (either needs to be supplied or otherwise risks performance penalty). Should be enough to allow a variant with IEqualityComparer<T> parameter and make something like `EqCompare<T>` public to allow on-the-fly creation of Comparers using equals+gethashcode.
3. Do you mind merging https://github.com/louthy/language-ext/pull/841 (which is related here / giving wrapper EQ => IEqualityComparer)? This would be a (semi) replacement for (1.)

PS: In case you like to stay with 'comparer func parameter': Maybe code (and bugfix) should be improved by moving `IfNone` and `Match` outside to avoid calling this within each call of of the comparer. Maybe `comparer` the should be non-optional (adding another signature without comparer) to clean up code.

### DIFF
--- a/LanguageExt.Core/Immutable Collections/Seq/Seq.Extensions.cs
+++ b/LanguageExt.Core/Immutable Collections/Seq/Seq.Extensions.cs
@@ -476,7 +476,7 @@ public static class SeqExtensions
     /// <returns>A new sequence with all duplicate values removed</returns>
     [Pure]
     public static Seq<T> Distinct<T, K>(this Seq<T> list, Func<T, K> keySelector, Option<Func<K, K, bool>> compare = default(Option<Func<K, K, bool>>)) =>
-        toSeq(Enumerable.Distinct(list, new EqCompare<T>((a, b) => compare.IfNone(default(EqDefault<K>).Equals)(keySelector(a), keySelector(b)), a => keySelector(a)?.GetHashCode() ?? 0)));
+        toSeq(Enumerable.Distinct(list, new EqCompare<T>((a, b) => compare.IfNone(default(EqDefault<K>).Equals)(keySelector(a), keySelector(b)), a => compare.Match(_  => 0, () => keySelector(a)?.GetHashCode() ?? 0))));
 
     /// <summary>
     /// Apply a sequence of values to a sequence of functions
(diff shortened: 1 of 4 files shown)


LanguageExt.Core/Immutable Collections/Seq/Seq.Extensions.cs
```diff
@@ -476,7 +476,7 @@ public static class SeqExtensions
     /// <returns>A new sequence with all duplicate values removed</returns>
     [Pure]
     public static Seq<T> Distinct<T, K>(this Seq<T> list, Func<T, K> keySelector, Option<Func<K, K, bool>> compare = default(Option<Func<K, K, bool>>)) =>
-        toSeq(Enumerable.Distinct(list, new EqCompare<T>((a, b) => compare.IfNone(default(EqDefault<K>).Equals)(keySelector(a), keySelector(b)), a => keySelector(a)?.GetHashCode() ?? 0)));
+        toSeq(Enumerable.Distinct(list, new EqCompare<T>((a, b) => compare.IfNone(default(EqDefault<K>).Equals)(keySelector(a), keySelector(b)), a => compare.Match(_  => 0, () => keySelector(a)?.GetHashCode() ?? 0))));
```
louthy, line 479: I think this would probably be clearer an the intent:
```c#
    public static Seq<T> Distinct<T, K>(this Seq<T> list, Func<T, K> keySelector, Option<Func<K, K, bool>> compare = default) =>
        toSeq(Enumerable.Distinct(list, 
            new EqCompare<T>(
                (a, b) => compare.IfNone(default(EqDefault<K>).Equals)(keySelector(a), keySelector(b)), 
                a => compare.Match(Some: _  => 0, None: () => default(EqDefault<K>).GetHashCode(keySelector(a))))));

```

louthy:
Thanks @StefanBertels 👍

PR #1083, opened 2022-07-11
Fix incorrect docstring for two LeftToSeq implementations.

ode2code95:
This is a minor correction of some incorrect doc strings that have tripped me up more than once. So I thought I'd contribute a fix.


louthy:
Thanks @ode2code95 👍

PR #1093, opened 2022-08-02
Microsoft.Bcl.AsyncInterfaces upgrade from 5.0.0 to 6.0.0

StanJav:
The commit [Break out transformer extension methods into their own package](https://github.com/louthy/language-ext/commit/d6fb39a34b973708ab955ce6e6c68bdfc9d2ad51) forces everybody on non-core FW to use `Microsoft.Bcl.AsyncInterfaces` v5 or the latest v6 but with an assembly redirect.

I am updating it to v6 so we are on the latest.


louthy:
Thanks Stan

PR #1100, opened 2022-08-11
Added Prism optics for manipulating Option<A> fields

twaddell:
Composing Lenses for records with optional fields can get a bit messy.
Haskell and other FP languages have the concept of Prisms for dealing with these cases.

A Prism is very similar to a Lens but with the following signatures for its Get and Set
 Get: `Func<A, Option<B>>`
 Set: `Func<B, Func<A, A>>`

The Get maps the option to the child properties attempting to be accessed and the Set will only set values if the Prism chain is all Somes.


louthy:
Looks good, thanks @twaddell 👍

PR #1114, opened 2022-09-21
Fix [Record] HashableAttribute (runtime error)

StefanBertels:
Using `[Hashable(...)]` in `[Record]` results in runtime error.

Found this bug because I didn't use `[Hashable(...)]` but just  `[Eq(...)]` and  `[Ord(...)]` resulting in different hash codes.

---

IMHO it's error-prone (and boilerplate) to require all three attributes for simple cases like the one shown in the test case.

GetHashCode maybe should be implemented by using the first existing of
1. [Hashable]
2. [Eq]
3. [Ord]
4. HashDefault

I'm not sure whether `Ord` should (or does) fully superseed `Eq` which could result in a similar list for implementation of `Equals`. I often use [Record] types having one or more case insensitive string properties as keys for e.g. `Map` and `HashMap` or with `.GroupBy(..)`.

Edit: somehow related to https://github.com/louthy/language-ext/pull/842 (CompareTo==0 vs. Equals==true)

### DIFF
--- a/LanguageExt.Core/Class Instances/Hashable/HashableException.cs
+++ b/LanguageExt.Core/Class Instances/Hashable/HashableException.cs
@@ -1,6 +1,7 @@
 using System;
 using System.Diagnostics.Contracts;
 using System.Threading.Tasks;
+using LanguageExt.TypeClasses;
 
 namespace LanguageExt.ClassInstances
 {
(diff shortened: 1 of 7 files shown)


louthy:

I thought I already fixed this, if you still care (I realise it's a bit late!) feel free to re-check and let me know.

Issue #1115, opened 2022-09-23
Map<K,V>.itemOrNone Lens throws exception if key doesn't exist in map


twaddell:
The behaviour of the `itemOrNone` Lens Get for a `Lst<A>` is to return a `None` if the index requested is out of range.
I expected the behaviour of `itemOrNone` Lens Get for a `Map<K,V>` to be similar in that if the key doesn't exist a `None` would be returned. However it currently has the same behaviour as the `item` Lens, in that it throws an Exception.

I think it would be an improvement to return `None` if the key doesn't exist.

This is also appears to be the case for `HashMap` and `TrackingHashMap`

PR #1116, opened 2022-09-23
Fix for #1115 Map<K,V>.itemOrNone Lens Get to return None when key does not exist

twaddell:
Updates the behaviour of itemOrNone Lenses for `Map`, `HashMap` and `TrackingHashMap` to return None if key doesn't exist in the map.
This resolves the issue raised in [#1115](https://github.com/louthy/language-ext/issues/1115)

Also includes unit tests to verify the new behaviour and the existing behaviour of `Arr` and `Lst`  itemOrNone Lenses


louthy:
Looks good @twaddell , Thanks 👍

PR #1134, opened 2022-10-26
Fix error message of notFollowedBy parser

StefanBertels:
1. [BUG] Tests shows that notFollowedBy reports wrong position. Fixed.
2. notFollowedBy did output result of p. This might be something different than input string and therefore might hide the problem. As `label` is not available I think it is a better approach to output the problematic part of the input string (at least if something was consumed).
3. The new version will allow parsers that don't consume anything like `eof`. Might be rare but why not support `notFollowedBy(eof)` or parsers that are decorated with `lookahead` etc.

Notes:
I tried handling all cases of ParserResult state correctly, but I'm not sure -- especially about the last case.

### DIFF
--- a/LanguageExt.Parsec/Parsers/Prim.cs
+++ b/LanguageExt.Parsec/Parsers/Prim.cs
@@ -751,11 +751,30 @@ namespace LanguageExt.Parsec
         ///                                
         /// </example>
         public static Parser<Unit> notFollowedBy<T>(Parser<T> p) =>
-            attempt(
-                either(from c in attempt(p)
-                       from u in unexpected<Unit>(c.ToString())
-                       select u,
-                       result(unit)));
+            inp =>
+            {
+                var res = p(inp);
+                if (res.Tag == ResultTag.Consumed)
+                {
+                    var message = inp.Index == res.Reply.State.Index - 1
+                        ? $"'{inp.Value[inp.Index]}'" // single char
+                        : $"\"{inp.Value.Substring(inp.Index, res.Reply.State.Index - inp.Index)}\""; // empty string or more than 1 char
+                    return EmptyError<Unit>(ParserError.Unexpect(inp.Pos, message));
+                }
+                if (res.Reply.Tag == ReplyTag.Error)
+                {
+                    return EmptyOK(unit, inp);
+                }
+                if (res.Reply.Error == null)
+                {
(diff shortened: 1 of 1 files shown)


louthy:
The wrong position reporting is a bug.  If you look at the Haskell implementation (that this is a clone of):

```haskell
notFollowedBy p = 
    try (do c <- try p
            unexpected (show c)) 
        <|> return ())
```
You can see how the C# version is the same:
```c#
attempt(
    either(from c in attempt(p)
           from u in unexpected<Unit>(c.ToString())
           select u,
           result(unit)));
```
(`try` is `attempt` and `<|>` is `either`)

It looks to me that either there's a bug in the Haskell original (probable, because it looks like the position would be updated to past the `p`).  A better way to do this without unpacking everything would be to use a bespoke `unexpected` function:

```c#
public static Parser<T> unexpected<T>(Pos pos, string msg) =>
    _ => EmptyError<T>(ParserError.Unexpect(pos, msg));
```
Combined with `getPos`:

```c#
attempt(
    either(from l in getPos
           from c in attempt(p)
           from u in unexpected<Unit>(l, c.ToString())
           select u,
           result(unit)));
```

Not sure I agree on point 2 and 3.  `notFollowedBy(eof)` doesn't really make sense.  Because, mostly we want to assert the end of the stream, not assert that we're _not_ at the end of the stream, i.e. "we expect to be finished and there shouldn't be anything left", not "have we got to the end of the file yet?".  So, for any Parser `p`, we should do:

```c#
var p1 = from r in p
         from _ in eof
         select r;
```

Let's say we were parsing lines in a CSV file:
```c#
var lines = from ls in many1(line)
            from _  in eos
            select ls;
```
In terms of what to show: stream tokens or parsed value.  The parsed value should be shown.  It may make sense for some circumstances to show the steam tokens, but the raw tokens aren't always `char` (`Parser<I, O>` and `Parser<A>` should work the same).  

I think I'd prefer to keep the combinator as it is.  And presumably by now you've written your own `notFollowedBy` that works how you want it?!

What I will do though, is update the `notFollowedBy` to allow for a bespoke label:
```c#
public static Parser<Unit> notFollowedBy<T>(Parser<T> p, string? label = null) =>
    attempt(
        either(from l in getPos
               from c in attempt(p)
               from u in unexpected<Unit>(l, label ?? c.ToString())
               select u,
               result(unit)));
```
I think that's a reasonable compromise.  So, I'll close this.

PR #1146, opened 2022-11-21
awaiting an OptionAsync should not throw ValueIsNoneException

BrunoJuchli:
Doing

    await OptionAsync<T>.None;

throws a `ValueIsNoneException` exception.

IMO this is not helpful
 - it's masking that we're working with an option type where we should expect/handle the none case
 - handling the none case with a try/catch seems a like a bad idea to me

Therefore I propose the awaiter to return an `Option<A>` (instead of `A`).

Note, that of course this is a **breaking change** to the API. Personally I don't really expect people have used it, but I don't know.

It's a simple change and it does not cause any of the unit tests to fail.

### DIFF
--- a/LanguageExt.Core/Monads/Alternative Value Monads/Option/OptionAsync/OptionAsync.Awaiter.cs
+++ b/LanguageExt.Core/Monads/Alternative Value Monads/Option/OptionAsync/OptionAsync.Awaiter.cs
@@ -16,12 +16,12 @@ namespace LanguageExt
         public bool IsCompleted =>
             awaiter.IsCompleted;
 
-        public A GetResult()
+        public Option<A> GetResult()
         {
             var (isSome, value) = awaiter.GetResult();
             return isSome
                 ? value
-                : throw new ValueIsNoneException();
+                : Option<A>.None;
         }
 
         public void OnCompleted(Action completion) =>
--- a/LanguageExt.Tests/OptionAsyncTests.cs
+++ b/LanguageExt.Tests/OptionAsyncTests.cs
@@ -300,6 +300,12 @@ namespace LanguageExt.Tests
             Assert.True(result);
         }
 
+        [Fact]
+        public async void AwaitAsync_IsNone()
+        {
+            await GetValue(false);
+        }
+
         // Not valuable any more 
         //[Fact]
         //public async void SequenceFlip()
--- a/LanguageExt.Tests/OptionAsyncTests.cs
+++ b/LanguageExt.Tests/OptionAsyncTests.cs
@@ -300,10 +300,14 @@ namespace LanguageExt.Tests
             Assert.True(result);
         }
 
-        [Fact]
-        public async void AwaitAsync_IsNone()
+        [Theory]
+        [InlineData(true)]
+        [InlineData(false)]
+        public async void AwaitAsync(bool isSome)
         {
-            await GetValue(false);
+            var result = await GetValue(isSome);
+            
+            Assert.Equal(isSome, result.IsSome);
         }
 
         // Not valuable any more 


louthy:
Await only has one shortcutting operation, which is exceptions, that's why this approach was taken - to use `await` like `<-` in Haskell, or `in` with LINQ.  For those wanting to use `OptionAsync` imperatively this implements that.  

However, [I have already removed this from version 5](https://github.com/louthy/language-ext/blob/version-five-refactor/LanguageExt.Core/Monads/Alternative%20Value%20Monads/Option/OptionAsync/OptionAsync.cs#L34), as I have come to the conclusion that this functionality is leaky at best and misleading at worst, especially when combining different awaiter types.  You should use LINQ for sequentional monadic operations instead.

BrunoJuchli:
@louthy 
Thank you for removing the problematic behavior.

TBH I don't understand the haskell reference, hopefully I will some day ;-)

Personally I would prefer `OptionAsync<A>` to behave similary to `EitherAsync<L,R>` which is awaitable (and returns an `Either<L,R>`).

Implementation and maintenance wise IMO this is trivial:

    public TaskAwaiter<Option<A>> GetAwaiter() =>
        ToOption().GetAwaiter();

But if you don't want to add that to the library it's easy for us to just add this via extension method:

    public static OptionAsyncExtensions
    {
        public static TaskAwaiter<Option<A>> GetAwaiter<A>(
            this OptionAsync<A> optionAsync) =>
            optionAsync.ToOption().GetAwaiter();
    }

(only downsides that this needs the `using` directive for the namespace of the extension method...)

louthy:
> Personally I would prefer OptionAsync<A> to behave similary to EitherAsync<L,R> which is awaitable (and returns an Either<L,R>).

You highlight exactly the problem.  It's entirely _opinion_ what should be returned from awaiting an `OptionAsync`, should it:

* Return an `Option<A>` - kinda makes sense, but then you have to unpack the `Option<A>`.  Which means a further `Match`, or a `switch` on the `.Case`.  This also isn't like `await` for 99.9% of everybody's expectations.  Most people expect a usable value to come out of `await`, not one that is still wrapped.
* Return a `Task<A?>` - this is the closed to returning an optional value, the closest to the expected _unpacking_ behaviour of `await`.  And with nullable references can leverage the machinery of the C# compiler to make pattern-matching relatively easy.  However `A?` isn't `Option<A>`:  nullable references aren't bulletproof and some users don't have then enabled.  At best this can be seen as a natural transformation from `Option<A>` to `A?
* Return a `Task<A>` - like above, but never return `null` or `None` - throw a `ValueIsNoneException`.  This leverages the `async / await` machinery to have a monadic shortcut that ends the process when we have no value.  This means no checking, switching, Matching is needed - the value is always good.  The downside is it uses exceptions, which need to be caught.  So at the top level you need to find a way of wrapping that back up in an `Option`.
* Return an `Either<Exception, A>` (or `Fin<A>`) - This allows the capture of any exception when awaiting the task.  Kinda makes sense, except that C# await already has exception handling.  It has many of the problems of awaiting an `Option<A>`.

Unltimately these are all compromised and nothing is clear from the type-signature of the `OptionAsync<A>` what behaviour you'll get.  By removing the `GetAwaiter` from `OptionAsync` I discourage the imperative nature of `async / await`, which is good, and I encourage the use of LINQ for working with monads, which _is_ declarative and expression based - all positive.

It also allow others to follow their own instincts on this.  There is clear disagreement on how this should be implemented, because it is ultimately just opinion in the end.  By removing it, I remove my opinion and open it up for you to have yours.  This library is usually super opinionated about functional concepts, but this isn't a functional concept, it's an imperative one (which is probably why it's not clear what the best approach is).

> (only downsides that this needs the using directive for the namespace of the extension method...)

You can declare your extension in the top-level (without a namesapce declaration), then it will always be included.

PR #1160, opened 2022-12-20
Add EitherDataExtensions

rammba:
Hello team, thanks for implementing this library. I recently started using it and I can say that it's really helpful.

Added few conversion extensions for `EitherData`:
- ToEither
- ToEitherAsync
- ToEitherUnsafe
- ToFin
- ToTry
- ToTryAsync

### DIFF
--- a/LanguageExt.Core/Monads/Alternative Value Monads/Either/Either-Shared/EitherData.Extensions.cs
+++ b/LanguageExt.Core/Monads/Alternative Value Monads/Either/Either-Shared/EitherData.Extensions.cs
@@ -1,0 +1,19 @@
+﻿using System;
+
+namespace LanguageExt.DataTypes.Serialisation
+{
+    public static class EitherDataExtensions
+    {
+        public static Either<L, R> ToEither<L, R>(this EitherData<L, R> input) =>
+            input.State switch
+            {
+                EitherStatus.IsBottom => Either<L, R>.Bottom,
+                EitherStatus.IsRight => Either<L, R>.Right(input.Right),
+                EitherStatus.IsLeft => Either<L, R>.Left(input.Left),
+                _ => throw new InvalidOperationException($"Invalid {nameof(EitherStatus)} with value: {input.State}"),
+            };
+
+        public static EitherAsync<L, R> ToEitherAsync<L, R>(this EitherData<L, R> input) =>
+            input.ToEither().ToAsync();
+    }
+}
--- a/LanguageExt.Core/Monads/Alternative Value Monads/Either/Either-Shared/EitherData.Extensions.cs
+++ b/LanguageExt.Core/Monads/Alternative Value Monads/Either/Either-Shared/EitherData.Extensions.cs
@@ -1,4 +1,6 @@
-﻿using System;
+﻿#nullable enable
+using System;
+using System.Threading.Tasks;
 
 namespace LanguageExt.DataTypes.Serialisation
 {
@@ -15,5 +17,11 @@ namespace LanguageExt.DataTypes.Serialisation
 
         public static EitherAsync<L, R> ToEitherAsync<L, R>(this EitherData<L, R> input) =>
             input.ToEither().ToAsync();
+
+        public static async Task<EitherAsync<L, R>> ToEitherAsync<L, R>(this Task<EitherData<L, R>> input) =>
+            (await input).ToEitherAsync();
+
+        public static async Task<EitherAsync<L, R>> ToEitherAsync<L, R>(this ValueTask<EitherData<L, R>> input) =>
+            (await input).ToEitherAsync();
     }
 }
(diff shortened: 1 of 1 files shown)


LanguageExt.Core/Monads/Alternative Value Monads/Either/Either-Shared/EitherData.Extensions.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System;
```
louthy, line 1: Could you add `#nullable enable` to the top of the file pls.  language-ext is being slowly migrated to support nullable references, it'd be good if new code supported it off the bat.

LanguageExt.Core/Monads/Alternative Value Monads/Either/Either-Shared/EitherData.Extensions.cs
```diff
@@ -0,0 +1,19 @@
...
+            input.State switch
+            {
+                EitherStatus.IsBottom => Either<L, R>.Bottom,
+                EitherStatus.IsRight => Either<L, R>.Right(input.Right),
+                EitherStatus.IsLeft => Either<L, R>.Left(input.Left),
+                _ => throw new InvalidOperationException($"Invalid {nameof(EitherStatus)} with value: {input.State}"),
+            };
+
+        public static EitherAsync<L, R> ToEitherAsync<L, R>(this EitherData<L, R> input) =>
+            input.ToEither().ToAsync();
```
louthy, line 17: There should also be a `ToEitherAsync` that takes `ValueTask<EitherData<L, R>>` and one that takes `Task<EitherData<L, R>>`

louthy (changes requested):
There's other types that should be supported:

* `EitherUnsafe<L, R>` (allows nulls in the `L` and `R`)
* `Fin<A>` which is the same as `EitherData<Error, A>`
* `Try<A>` which is the same as `EitherData<Exception, A>`
* `TryAsync<A>` which is the same as `Task<EitherData<Exception, A>>` and `ValueTask<EitherData<Exception, A>>` 

Welcome to my world.  There's no such thing as a small change in this library ;-)

rammba:
> There's other types that should be supported:
> 
> * `EitherUnsafe<L, R>` (allows nulls in the `L` and `R`)
> * `Fin<A>` which is the same as `EitherData<Error, A>`
> * `Try<A>` which is the same as `EitherData<Exception, A>`
> * `TryAsync<A>` which is the same as `Task<EitherData<Exception, A>>` and `ValueTask<EitherData<Exception, A>>`
> 
> Welcome to my world. There's no such thing as a small change in this library ;-)

Thanks for drawing attention on other monads, I hope my conversion is correct regarding your implementation.

LanguageExt.Core/Monads/Alternative Value Monads/Either/Either-Shared/EitherData.Extensions.cs
```diff
@@ -0,0 +1,64 @@
...
+{
+    public static class EitherDataExtensions
+    {
+        public static Either<L, R> ToEither<L, R>(this EitherData<L, R> input) =>
+            input.State switch
+            {
+                EitherStatus.IsBottom => Either<L, R>.Bottom,
+                EitherStatus.IsRight => Either<L, R>.Right(input.Right),
+                EitherStatus.IsLeft => Either<L, R>.Left(input.Left),
+                _ => throw new InvalidOperationException($"Invalid {nameof(EitherStatus)} with value: {input.State}"),
```
rammba, line 16: Is this correct approach with `InvalidOperationException`, or to use something like

```suggestion
                EitherStatus.IsRight => Either<L, R>.Right(input.Right),
                EitherStatus.IsLeft => Either<L, R>.Left(input.Left),
                _ => Either<L, R>.Bottom,
```

louthy: I think the exception approach is best here, as it is an exceptional and unknown value, defaulting to bottom is misleading I think.

LanguageExt.Core/Monads/Alternative Value Monads/Either/Either-Shared/EitherData.Extensions.cs
```diff
@@ -0,0 +1,64 @@
...
+            };
+
+        public static TryAsync<A> ToTryAsync<E, A>(this EitherData<E, A> input) where E : Exception =>
+            input.ToTry().ToAsync();
+
+        public static async Task<TryAsync<A>> ToTryAsync<E, A>(this Task<EitherData<E, A>> input) where E : Exception =>
+            (await input).ToTryAsync();
+
+        public static async Task<TryAsync<A>> ToTryAsync<E, A>(this ValueTask<EitherData<E, A>> input) where E : Exception =>
+            (await input).ToTryAsync();
```
louthy, line 62: `.ConfigureAwait(false)`

louthy (changes requested):
Almost there, just a few extras needed.

louthy:
Looks good @rammba , Thanks! 👍

PR #1168, opened 2023-01-10
Fix Map<Ord,...> + Map<Ord,..> result type

StefanBertels:
`Map<OrdK,...> + Map<OrdK,..>` returned a Map type without `OrdK`


louthy:
Thanks @StefanBertels 👍

PR #1169, opened 2023-01-10
Remove duplicate assignment in PString constructor

timmi-on-rails:


louthy:
Thanks @timmi-on-rails 👍

PR #1180, opened 2023-01-29
Fix race condition when merging producers

### LINKED ISSUE #1179: Merge synchronous producers causes exception
Consider the following effect, where the `Runtime` is arbitrary:

```csharp
Effect<Runtime, Unit> effect = compose(
    Producer.merge<Runtime, int, Unit>(yield(1), yield(1)),
    awaiting<int>().Map(ignore))
```

I would expect the execution of this effect to succeed.
But what happens is, an exception is thrown: `LanguageExt.Common.ExpectedException : sequence empty`.

The problem seems related to all producers being synchronous.
My proposed solution can be found in the linked PR.

timmi-on-rails:
The resulting producer of the function `Producer.merge` failed with an exception, when only synchronous producers were combined. The boolean variable `running` was set to false, before any items of the `queue` were yielded.

Fixes #1179

### DIFF
--- a/LanguageExt.Core/Effects/Pipes/Producer/Producer.cs
+++ b/LanguageExt.Core/Effects/Pipes/Producer/Producer.cs
@@ -428,8 +428,7 @@ namespace LanguageExt.Pipes
                           });
                 #pragma warning restore CS4014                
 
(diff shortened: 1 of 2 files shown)


louthy:
Looks like a nice and simple fix, thanks @timmi-on-rails 👍

PR #1189, opened 2023-02-08
Add reset color function to console trait

timmi-on-rails:
Regarding console color handling, this is the only missing function from the BCL. Thus this commit completes the console color functionalities.


timmi-on-rails:
The `MemoryConsole.ResetColor` function is not atomic which yields inconsistent states in concurrent scenarios.

Is `MemoryConsole` supposed to be thread-safe?

louthy:
Thanks @timmi-on-rails 👍

PR #1190, opened 2023-02-14
fix: value is null exception on activity span

### LINKED ISSUE #1162: SysX Activity.span fails with value is null exception
Any usage of Activity.span will throw a value is null exception.

See attached branch and tests for example. https://github.com/louthy/language-ext/compare/main...bmazzarol:language-ext:bugfix/activity_issues

Tests showing the issues 

https://github.com/bmazzarol/language-ext/blob/bugfix/activity_issues/LanguageExt.Tests/ActivityTests.cs
https://github.com/bmazzarol/language-ext/blob/bugfix/activity_issues/LanguageExt.Tests/FinNullabilityIssueTests.cs

Working Activity (My copy)

https://github.com/bmazzarol/language-ext/blob/bugfix/activity_issues/LanguageExt.SysX/SysX/Diag/Activity2.cs

There are a few issues,

* **Fin is not nullable aware:** If you create an Aff or Eff with a type signature containing a nullable T, if the value is actually null, it will fail. This is the case for activity, the spec is Eff\<RT,Activity?\>. See FinNullabilityIssueTests.cs
* **default access to runtime:** Its only safe to access the runtime via default, if the value/operation you need does not require access to an inner environment class (user provided state that can only exist when the runtime is provided to the effect). In the case of ActivitySourceIO it requires a provided ActivitySource, which can never exist in a call to default\<RT\>. IMHO its not a good idea to access the runtime in this way, and it should be recommended to just create a new Eff or Aff or just use runtime\<RT\>. This way you are always guaranteed to be accessing the correct instance of the runtime. Moreover if a developer refactors the runtime code, they dont need to worry about all call sites and how the runtime is accessed. 

Starting this issue to talk through ways to resolve the follow,

* **Advice around nullable types and there use in monadic types:** I know that nullable reference types are nothing but compiler magic, but in net6 + repos that is the default and I have seen developers use it in code with similar effect to Fin, it was not an ideal experience. At the very least Fin needs a not null type constraint on its T
* **Advice around runtime usage:** the docs show accessing the runtime using default. I think that should be reconsidered, 
  * Its not safe in all cases
  * Its an additional concept that developers need to learn, (adhoc polymorphism is not required when I teach FP to developers)
  * The alternative runtime\<RT\> is cleaner IMHO
* **Fix SysX OTEL features:** would prefer to not have duplicated versions of LanguageExt code in my libraries.

bmazzarol:
fixes #1162

Fin does not allow null values, if the activity was null the Map on activity source would fail.

This wraps the potentially nullable activity in a wrapper that is never null.

I have also,

* added a missing operation addEvent
* added tests to all operations

### DIFF
--- a/LanguageExt.SysX/SysX/Diag/Activity.cs
+++ b/LanguageExt.SysX/SysX/Diag/Activity.cs
@@ -13,199 +13,221 @@ namespace LanguageExt.SysX.Diag
     /// takes an `Eff` or `Aff` operation to run (which is the activity).  The runtime system will maintain the parent-
     /// child relationships for the activities, and maintains the 'current' activity.
     /// </summary>
-    /// <typeparam name="RT"></typeparam>
+    /// <typeparam name="RT">runtime</typeparam>
     public static class Activity<RT>
         where RT : struct, HasActivitySource<RT>, HasCancel<RT>
     {
+        private readonly record struct DisposableActivity : IDisposable
+        {
+            public DisposableActivity(Activity? activity) => this.Activity = activity;
+
+            public void Dispose() => Activity?.Dispose();
+
+            public Activity? Activity { get; }
+        }
+
+        private static Eff<RT, DisposableActivity> startActivity(
+            string name,
+            ActivityKind activityKind,
+            HashMap<string, object> activityTags,
+            Seq<ActivityLink> activityLinks,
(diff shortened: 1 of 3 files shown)


louthy:
Thanks @bmazzarol 👍

PR #1194, opened 2023-02-19
feat: Support for Random IO

bmazzarol:
This implements an environment trait to support injectable random IO.

This can be extended to support new performance improvements in net6+ around generating fractional primitives.

This does not re-invent anything, it just wraps the default netcore Random class and adds enough functions on top to be useful. Inspiration from the Haskell and Elm Random modules was taken, but no additional type-class(es) was added, this can be looked into later, if the flexibility offered here is not enough.

Another great random module is the rust one, that was very powerful but overkill for what most C# dev needs.

Open to any and all comments and suggestions for improvement.

### DIFF
--- a/LanguageExt.Sys/Live/Runtime.cs
+++ b/LanguageExt.Sys/Live/Runtime.cs
@@ -1,3 +1,5 @@
+#nullable enable
+
 using System;
 using System.Text;
 using System.Threading;
@@ -19,7 +21,8 @@ namespace LanguageExt.Sys.Live
         HasTextRead<Runtime>,
         HasTime<Runtime>,
         HasEnvironment<Runtime>,
-        HasDirectory<Runtime>
+        HasDirectory<Runtime>,
+        HasRandom<Runtime>
     {
         readonly RuntimeEnv env;
 
@@ -133,6 +136,13 @@ namespace LanguageExt.Sys.Live
         /// <returns>Operating-system environment environment</returns>
         public Eff<Runtime, Traits.EnvironmentIO> EnvironmentEff =>
             SuccessEff(Sys.Live.EnvironmentIO.Default);
+        
+        /// <summary>
+        /// Access the random synchronous effect environment
+        /// </summary>
+        /// <returns>Random synchronous effect environment</returns>
+        public Eff<Runtime, Traits.RandomIO> RandomEff =>
+            SuccessEff<Runtime, Traits.RandomIO>(new Sys.Live.RandomIO(new Random()));
     }
     
     public class RuntimeEnv
--- a/LanguageExt.Sys/Live/Runtime.cs
+++ b/LanguageExt.Sys/Live/Runtime.cs
@@ -41,30 +41,34 @@ namespace LanguageExt.Sys.Live
         /// <summary>
         /// Constructor function
         /// </summary>
-        public static Runtime New() =>
-            new Runtime(new RuntimeEnv(new CancellationTokenSource(), System.Text.Encoding.Default));
+        /// <param name="seed">seed to used for the random generator</param>
+        public static Runtime New(int? seed = default) =>
+            new Runtime(new RuntimeEnv(new CancellationTokenSource(), Encoding.Default, seed));
 
         /// <summary>
         /// Constructor function
         /// </summary>
         /// <param name="source">Cancellation token source</param>
-        public static Runtime New(CancellationTokenSource source) =>
-            new Runtime(new RuntimeEnv(source, System.Text.Encoding.Default));
+        /// <param name="seed">seed to used for the random generator</param>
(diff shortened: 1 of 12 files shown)


LanguageExt.Sys/Live/Runtime.cs
```diff
@@ -133,6 +136,13 @@ namespace LanguageExt.Sys.Live
         /// <returns>Operating-system environment environment</returns>
         public Eff<Runtime, Traits.EnvironmentIO> EnvironmentEff =>
             SuccessEff(Sys.Live.EnvironmentIO.Default);
+        
+        /// <summary>
+        /// Access the random synchronous effect environment
+        /// </summary>
+        /// <returns>Random synchronous effect environment</returns>
+        public Eff<Runtime, Traits.RandomIO> RandomEff =>
+            SuccessEff<Runtime, Traits.RandomIO>(new Sys.Live.RandomIO(new Random()));
```
timmi-on-rails, line 145: Creating a new random generator here, means you will use a new random generator for each line in the following example:

```csharp
from x in Random<RT>.nextInt()
from y in Random<RT>.nextInt()
from z in Random<RT>.nextInt()
select x + y + z
```

When I want to use the same random generator I can't use the static `Runtime<RT>` helper methods.
That would be sad.

I don't have a nice solution, but here is an idea:

The live runtime holds the `Random` instance and the seed in its `RuntimeEnv`.
The live `RandomIO` synchronizes the access to the random generator from multiple threads. (e.g. https://learn.microsoft.com/en-us/dotnet/api/system.random?view=netstandard-2.1#the-systemrandom-class-and-thread-safety).
Similar to the `LocalCancel` property of the `HasCancel<RT>` trait, we add a `LocalRandom` property which allows for a new random generator in case I don't want to use the runtime global one.
Maybe `LocalRandom` should be a method, where I can use a different seed.

bmazzarol: Thanks for review!

I agree, will make some changes there,

* I expect 1 random instance per runtime
* A local random with optional seed would be a great feature

bmazzarol: @timmi-on-rails have made the changes so that Random is scoped to the runtime instance.

Not 100% on-board with the recommended approach outlined above, need to talk through the benefits of a singleton source of randomness, as apposed to a random instance scoped to the runtime.

timmi-on-rails: I didn't mean to create a singleton.
One Random instance per runtime was my intention as well.
Yet, you need to address thread- safety.
I will sight your changes later.

bmazzarol: All good, I  understand.

Might start by creating a test that corrupts the Random instance and try a few synchronization approaches. 

* The recommended approach with locks
* The Random instance per thread local approach
* Some sort of interlock variation. Not going lie, low level concurrency is not my strongest skill

The goals,

* No corrupt Random, users should not get streams of zeros
* Support for user seeds, and local seeds.
* Performance is not compromised

bmazzarol: @timmi-on-rails random IO is now fully thread safe and the live instance is singleton scoped to each thread that accesses it (using thread local).

Should provide for the best of both worlds.

Deterministic random generators for tests, correct unbiased and fast random generation live.

User provided seeds for the Random generator are supported, the live instance memoizes the thread local instance for all unique seeds used.

The risk here is that the user leaks memory by randomly generating seeds, If that is done, the user is on their own.

The case would be a misuse of seeds and would be pointless IMHO.

If you are happy you can resolve it.

timmi-on-rails: Why did you go for a static Random instance per thread?
I would prefer runtime-scoped Random instances for better isolation.

bmazzarol: Because it generates more uniform sequences and is lock free.

If you have a 100% pure application it will effectively be a singleton, as there is a single entry point. However it's not thread safe so locks are required.

But consider many hybrid codebases with many entry points, the runtime does not live very long. So might cause sequence issues and still requires locks.

With thread scoped singleton random instances it complies with the recommended approach outlined by MS, generates the most uniform sequences under any codebase configuration and is lock free.

Happy to consider counter arguments.

timmi-on-rails: The point about short-lived runtimes convinced me, e.g. runtime per request in a webserver.

(BTW: Thanks and I will answer your comment on my EF Core thread. Currently too busy.)

timmi-on-rails (approved):

bmazzarol:
Closing due to lack of interest, can consider opening again if its needed by anyone.

Issue #1193, opened 2023-02-17
`Prelude.OptionAsync` takes `T` instead of `T?`


BrunoJuchli:
These two overloads take a `Task<T>` instead of a `Task<T?>`:

 - `public static OptionAsync<T> OptionalAsync<T>(Task<T> value) =>`
https://github.com/louthy/language-ext/blob/cd07f2b49a4272eb313f375e1aab9034d9ed13c8/LanguageExt.Core/Monads/Alternative%20Value%20Monads/Option/OptionAsync/OptionAsync.Prelude.cs#L153
 - `public static OptionAsync<T> OptionalAsync<T>(Func<Unit, Task<T>> f) =>`
 https://github.com/louthy/language-ext/blob/cd07f2b49a4272eb313f375e1aab9034d9ed13c8/LanguageExt.Core/Monads/Alternative%20Value%20Monads/Option/OptionAsync/OptionAsync.Prelude.cs#L163
 
 
 ## Workaround
 Just use `OptionAsync<StoredOAuthApplication>.OptionalAsync(..)` directly.

louthy:
Those files haven't been converted to `#nullable enable` yet, so the `T` can be `null` for reference-types.

BrunoJuchli:
Thank you for the quick feedback.

My IDE thinks otherwise:

![image](https://user-images.githubusercontent.com/2090172/220016349-d3b3dd32-8638-4d9a-af2c-d8f988ee412e.png)

Decompiling the file shows:
![image](https://user-images.githubusercontent.com/2090172/220016446-94dfe15c-179d-44de-8428-80fea438493a.png)

i'm using
- LanguageExt 4.4.2
- Rider 2022.3.2 (MsBuild 17)

louthy:
Ah, must've been looking at the wrong file!  Feel free to push a MR

BrunoJuchli:
In the process of creating a PR for this I noticed that you were right -- nullable isn't enabled on the (partial class) file.
I also looked at the IL and now I think it's a bug in Rider.

The produced IL looks like this:
![grafik](https://user-images.githubusercontent.com/2090172/221396512-f5a8137e-4397-4822-8056-c588dd85d8d4.png)

I went ahead and set `#nullable enable` in the top of OptionAsync.Prelude.cs and changed to `T?` as needed:
![grafik](https://user-images.githubusercontent.com/2090172/221396561-21d51d58-bc7e-4e96-b0cb-0d1cb36db9b5.png)

[...]

PR #1200, opened 2023-02-26
Nullability Annotations for Prelude of OptionAsync

BrunoJuchli:
This relates to #1193 - as [commented](https://github.com/louthy/language-ext/issues/1193#issuecomment-1445282575) this doesn't fix a bug in LanugageExt but just improves the API:

I enabled nullable for the whole of OptionAsync.Prelude.cs - went through all methods and changed nullability where necessary.

(I also ran all the unit tests which are still green).


louthy:
Thanks @timmi-on-rails 👍

Issue #1024, opened 2022-04-24
Arithmetic instances in LangaugeExt.TypeClass


benjstephenson:
I'm pretty new to the library (really loving it 👌) and C# in general but I've come from a Scala Cats etc world so I'm not sure if I'm just trying to overlay what I've learned in that space too much and getting myself confused a bit.

I can see that `Arithmetic` defines the `Plus`, `Subtract`, `Negate` and `Product` capabilities that `Num` extends from, I've got a custom data type that makes sense to have these implemented but not the rest of the methods that `Num` defines.
I've tried to use the `subtract` etc static methods from `Prelude` but found that they expect a null `Num` instance:

```csharp
public static A subtract<NUM, A>(A x, A y) where NUM : struct, Num<A> => default (NUM).Subtract(x, y);
```

Should these methods accept an instance of `Arithmetic` as I think that's the "least powerful" typeclass that could be used?

louthy:
> but found that they expect a null Num instance

This sentence ^ confused me a bit, because instances shouldn't ever be `null`.  So, I'll walk through how it works in C# in case there's a gap in your understanding.  Apologies if I'm explaining what you already know....

C# doesn't have 'official' support for ad-hoc polymorphism (other than the static interface methods in the latest C#, but that hasn't any backward compatibility).  However, there's a way to implement it by leveraging the fact that `struct` cannot be `null` (combined with constraints).

So, for example - this is a 'trait' or 'type-class' defined as an interface (sticking to monoid, as it's very simple):

```c#
    public interface Monoid<A>
    {
        A Empty();
        A Append(A x, A y);
    }
```

Then the 'instance' can be implemented as a `struct`:
```c#
    public struct MSeq<A> : Monoid<Seq<A>>
    {
        public Seq<A> Empty() => Seq<A>();
        public Seq<A> Append(Seq<A> x, Seq<A> y) => x + y;
    }
```

If I do this:
```c#
    var mx = default(MSeq<int>);
```
Then `mx` won't be `null` (because `MSeq` is a `struct`).  And so I can then call the methods on it:
```c#
    var xs = mx.Empty();
    var ys = mx.Append(xs, Seq(1, 2, 3));
```
We don't ever pass instances around as data, we just need to provide them as generic arguments and constrain them.  Here's a general example of `Append` (`mappend` in Haskell)

```c#
    public static A Append<MA, A>(A xs, A ys) where MA : struct, Monoid<A> => 
        default(MA).Append(xs, ys);
[...]

PR #1208, opened 2023-03-15
Add Prelude functions for Arithmetic typeclass

benjstephenson:
Based on the discussion in #1024 (better 1 year late than never!).
Moved the functions from `Num.Prelude` to `Arithmetic.Prelude` that relate to the Arithmetic typeclass and updated the usage in Option/Async Try/Async extensions.


louthy:
Thanks @benjstephenson 👍

PR #1223, opened 2023-05-18
Added support for simpler if fail to Try monads

mark-pro:
Added simpler support to try and try options monads and their cooresponding async computations

Added fluent IfFail extensions for
- Try
- TryAsync
- TryOption
- TryOptionAsync

Added static function ifFail for

- Try
- TryAsync
- TryOption
- TryOptionAsync


louthy:
Thanks @mark-pro 👍

PR #1241, opened 2023-08-19
Remove MemoryFS dependency on Windows Path implementation

gderham:
Hi, long term fan of language-ext 😄 

I ran all `LanguageExt.Tests` tests on macOS (.NET 7.0) and some fail, mostly due to `MemoryFS` using `Path.IsPathRooted()`, `Path.DirectorySeparatorChar` etc with error:

```
System.IO.IOException : Path not rooted: C:
   at LanguageExt.Sys.MemoryFS.ParsePath(String path) in /Users/_/code/language-ext/LanguageExt.Sys/MemoryFS.cs:line 51
```

In this PR I've modified `MemoryFS` and `MemoryFSTests` to behave the same independent of the host OS, and therefore enable the tests to pass. The replacement methods aren't 100% realistic but I hope acceptable for this usage. There remains a dependency on `System.IO` for some exceptions and other types.

`MemoryFS` use of  Windows file paths is unchanged so there is no breaking change.

Thanks

### DIFF
--- a/LanguageExt.Sys/MemoryFS.cs
+++ b/LanguageExt.Sys/MemoryFS.cs
@@ -26,13 +26,14 @@ namespace LanguageExt.Sys
     public class MemoryFS
     {
         readonly Atom<Entry> machine = Atom<Entry>(new FolderEntry("[machine]", DateTime.MinValue, DateTime.MinValue, DateTime.MinValue, default));
-        readonly static char[] invalidPath = Path.GetInvalidPathChars();
-        readonly static char[] invalidFile = Path.GetInvalidFileNameChars();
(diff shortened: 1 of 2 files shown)


louthy:
Thanks for this, I'll have a look in more detail next week.  My initial thoughts on a quick scan through is that probably `LanguageExt.Sys` needs its own `Path`  trait like the other traits (with platform specific implementations).  

I had naively thought that `Path` wouldn't have side-effects, I should never have given MS the benefit of that doubt!

louthy:
I'm closing PRs that have conflicts because of the major v5 refactor.  I have been developing primarily on Linux and have already updated the `MemoryFS`.  The tests pass, so should be good.  Sorry this was a bit late, thanks for the effort!

PR #1247, opened 2023-09-07
Fix fin enumerator

StefanBertels:
`Fin<T>` has wrong type in `IEnumerable<>`.

Side effect of the bug: XUnit will run into stack overflow if `Assert.Equal()` is used (see test).

Workaround: convert to `Option<T>` with e.g. `myFin.ToOption()`

### DIFF
--- a/LanguageExt.Core/Monads/Alternative Value Monads/Fin/Fin.cs
+++ b/LanguageExt.Core/Monads/Alternative Value Monads/Fin/Fin.cs
@@ -21,7 +21,7 @@ namespace LanguageExt
         IComparable<Fin<A>>, 
         IComparable, 
         IEquatable<Fin<A>>,
-        IEnumerable<Fin<A>>,
+        IEnumerable<A>,
         ISerializable
     {
         internal readonly Error error;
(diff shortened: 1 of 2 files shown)


louthy:
This probably needs a wider discussion.  Because `IEnumerable<>` was only there to facilitate serialisation; if you look at [`Either<L, R>`](https://github.com/louthy/language-ext/blob/main/LanguageExt.Core/Monads/Alternative%20Value%20Monads/Either/Either/Either.cs#L39), you'll see that its `IEnumerable<>` bound-type is `EitherData<L, R>`.  Again, this is because of the limitations of Json.NET - it was the only way I could reliably serialise without writing bespoke JsonConvertors (which I didn't want to do, because I didn't want to bring Json.NET into the `Core`).

I have however been investigating the idea of building bespoke `LanguageExt.Serialisation.*` libraries (where `*` would be `Newtonsoft`, `System.Json`, `XmlDocument`, `Protobuf`, etc.  Which means there'd be no need for `IEnumerable` implementations on `Option`, `Either`, `Fin`, etc.  Backwards compatibility could be maintained through some bespoke `SelectMany` operators, or just an expectation that conversion should be explicit, using `AsEnumerable()`.

So, at the moment, your fix _does fix_ an issue (honestly, I'm not sure what I was thinking when I wrote that!) - but it also makes it inconsistent with `Either` - probably, it should be `IEnumerable<EitherData<Error, A>>`.  But, that whole line of reasoning will go away in about a month when I start building the new serialisation libraries!

Maybe just use `Fin<A>.ToArray()` temporarily, whilst I have a think about the best course of action.  

For all: Please let me know if you have a strong opinion on this.

louthy:
I'm closing PRs that have conflicts because of the major v5 refactor.  In this case `Fin` has already been refactored to use `IEnumerable<A>`.  Serialisation isn't supported by the default library and will be provided by additional (serialiser specific) libraries, as per the roadmap.

PR #1249, opened 2023-09-12
Fix codegen record lens

StefanBertels:
Files containing '[Record]' classes (codegen) without `using LanguageExt;` failed to compile.


louthy:
I've accepted this even though I'm not supporting `CodeGen` any more.  It might be useful to others.

PR #1255, opened 2023-10-01
Add await support for `OptionAsync`

timmi-on-rails:
I am not sure, if there is a reason why `await EitherAsync<string, int>.Right(4)` should work
 and `await OptionAsync<int>.Some(4)` should not, but here is the fix.


louthy:
I'm closing PRs that have conflicts because of the major v5 refactor.  `OptionAsync` has been removed in `v5` and is now `OptionT<IO, A>`.

PR #1284, opened 2023-11-25
Fix AffRxExtensions.Consume blocking on stream completion

### LINKED ISSUE #1283: AffRxExtensions.Consume blocks the waiting thread on IObservable's completion
I'm learning the ways how to handle observables with this library. This is an example I get a problem with:
```cs
using System.Reactive.Linq;
using LanguageExt;
using LanguageExt.Sys;
using LanguageExt.Sys.Live;
// Generates numbers from 0 to 4 each 500ms
var obs =
    Observable.Generate(
        initialState: 0,
        condition: x => x < 5,
        iterate: x => x + 1,
        resultSelector: x => x,
        timeSelector: _ => TimeSpan.FromSeconds(0.5));
await obs.Consume(x => Console<Runtime>.writeLine($"{x}").ToAff()).Run(Runtime.New());
```
**Output**: "01234" with a blocked console window, when expected a closed one.

It seems that [AutoResetEvent.WaitOne()](https://github.com/louthy/language-ext/blob/5be0222eb1f8ebaa3839ecb2fef11645e5b8f709/LanguageExt.Rx/Aff.Extensions.cs#L37) blocks the thread forever, as there's no `onCompleted` handler passed when subscribing to the observable.
If it's just me doing something wrong, then I would be glad to be pointed to how to work with observables (specifically, the ones created from C# `event`s to handle them when they are fired).

aloslider:
Fixes #1283 
Adds **onCompleted** handler to `IObservable `and bool tracking flag to exit the loop if stream is completed.

### DIFF
--- a/LanguageExt.Rx/Aff.Extensions.cs
+++ b/LanguageExt.Rx/Aff.Extensions.cs
@@ -24,17 +24,24 @@ namespace LanguageExt
                 {
                     A   nextValue = default;
                     using var wait = new AutoResetEvent(false);
+                    bool isCompleted = false;
 
                     using var disp = ma.Subscribe(
                         (A x) =>
                         {
                             nextValue = x;
                             wait.Set();
+                        },
+                        () =>
+                        {
+                            isCompleted = true;
+                            wait.Set();
                         });
 
                     while (token == default || !token.IsCancellationRequested)
                     {
                         wait.WaitOne();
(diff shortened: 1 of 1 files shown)


louthy:

In this case `Aff` doesn't exist any more and the `Eff` version has already been fixed.

PR #1295, opened 2024-02-09
refactor: use global usings static for prelude

riezebosch:
not a big one, but nice to have prelude everywhere in your project with the (not so new) `global using static` feature from C# 10.


louthy:

As a side note, it's always worth asking me before doing any PR whether I'm likely to accept, or not (in this case, probably, but I hadn't considered this an issue that needed resolving).

PR #1318, opened 2024-04-21
Fix typo in readme

hlaueriksson:
An extra backtick broke the inline code formating.


hlaueriksson:
This was removed in b7fd9674079bf3e8ffe316640660fde46c852753

PR #1334, opened 2024-06-06
Fix issue where TraverseSerial() would eagerly start EitherAsync tasks

Liversage:
The TraverseSerial() method no longer prematurely starts EitherAsync tasks in a lazy Seq when preallocating an array for the result.


louthy:
@Liversage  This is now live.  Thanks for the contribution!  👍

Liversage:
> @Liversage This is now live. Thanks for the contribution! 👍

Thanks for creating this library!

PR #1170, opened 2023-01-11
Add Knot and NauticalMile

rammba:

PR #1369, opened 2024-09-19
Add Knot and NauticalMile

rammba:
Hi @louthy, thanks for letting me know that some files were moved. Here are the same changes as in #1170, but without conflicts.


louthy:
Thanks! All merged :+1:

PR #1376, opened 2024-10-02
Identify and Fix Issue with Error Monoid Implementation

micmarsh:
I noticed that the `Error`s `Combine` method always returns a `ManyErrors`, even if you're concatenating a single error and an `Empty` error. This violates the identity monoid laws (assuming we want to use `==`  rather than `Error.Is` to determine Error equality, which may not be true) in addition to just _feeling_ like a bug, at least to me.

Obviously, if this was intended or is fine, please disregard!


louthy:
Looks good, thanks!

Issue #1406, opened 2024-11-07
XML comment for Fin doesn't load properly


guythetechie:
The XML summary comment for the `Fin` class doesn't load properly.
![image](https://github.com/user-attachments/assets/590858f1-2b7f-4403-a75e-ac3d790bae14)

I think it's because of the unescaped `<` and `>` here:
https://github.com/louthy/language-ext/blob/abf61d3f0eb4fc8a2b46291c118df7eee2e2bf76/LanguageExt.Core/Monads/Alternative%20Monads/Fin/Fin.cs#L14

If I change it to the code below, it seems to work properly.
```csharp
/// <summary>
/// Equivalent of `Either&lt;Error, A&gt;`
/// Called `Fin` because it is expected to be used as the concrete result of a computation
/// </summary>
public class FinTest
...
```
![image](https://github.com/user-attachments/assets/846fe8a6-1992-4985-abf8-bade39ebab36)

### Who cares?
It might seem like a minor issue, but I plan on using `Fin`, `Either`, et. al. as Trojan horses for functional programming in codebases where maintainers aren't familiar with LanguageExt. `Either<Error, int>` is a very descriptive return type. `Fin<int>` is extremely convenient, but its meaning is less clear without mousing over the type for its description.

louthy:
https://github.com/louthy/language-ext/releases/tag/v5.0.0-beta-40

PR #1416, opened 2024-11-22
Fix  for "XML comment for Fin doesn't load properly #1406"

micmarsh:
https://github.com/louthy/language-ext/issues/1406


louthy:
@micmarsh  Thanks for this, but I decided to take [another approach](https://github.com/louthy/language-ext/releases/tag/v5.0.0-beta-40).

micmarsh:
Makes sense, glad the issue is addressed one way or the other!

PR #1450, opened 2025-03-10
Added NotNullWhenAttribute in isEmpty method

HernanFAR:
Is just that. For the C# compiler to know that the string instance is not when when using in ifs.


louthy:
Thanks :+1:

PR #1489, opened 2025-09-29
fix SourceT Skip and Take

beezee:
Thank you @louthy for this awesome library - it has made c sharp a viable language for me :)

I have been spending some time with the Streaming lib, and found a few of rough edges. This PR contains a surgical tweak to alleviate two of them.

If you look at this fiddle, you'll see how SourceT.Skip is currently broken: https://dotnetfiddle.net/l0jrZ6

It appears due to the lifecycle of a reducer the `amount` var as defined never lives to decrement.

If you look at this fiddle, you'll see how SourceT.Take is currently broken: https://dotnetfiddle.net/9PVwD9

Explanation is similar as to the situation with Skip, with one caveat.

From what I can tell, it looks like SourceT has no capacity to signal termination, so this fix makes "observable" Take correct, but the underlying source always runs til exhausted. 

This means that `forever` and variants are only usable if you really mean "forever".

I have been using an unfold pattern that I've built up over here: https://github.com/beezee/UnfoldingStreams/blob/main/UnfoldingStreams.Tests/Tests.cs

I've also implemented schedule support against that abstraction since I noticed you have that currently marked as TODO.

If you're interested in incorporating this approach, or would be interested in help adjusting the current representation to support termination (and scheduling) - I'd love to contribute towards that!

### DIFF
--- a/LanguageExt.Streaming/Transducers/TransducerM/DSL/SkipTransducer.cs
+++ b/LanguageExt.Streaming/Transducers/TransducerM/DSL/SkipTransducer.cs
@@ -5,9 +5,9 @@ namespace LanguageExt;
 record SkipTransducerM<M, A>(int Amount) : TransducerM<M, A, A>
     where M : Applicative<M>
 {
+    int amount = Amount;
     public override ReducerM<M, A, S> Reduce<S>(ReducerM<M, A, S> reducer)  
     {
-        var amount = Amount;
         return (s, x) =>
                {
                    if (amount > 0)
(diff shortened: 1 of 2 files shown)


louthy:
Just FYI, I have seen this, but I'm mid house move atm, so everything is up in the air. I expect to be back to full language-ext duty in a few weeks.

beezee:
Understood. I'll watch the new issue to see what you come up with and in the meantime I'll publish my unfold library to nuget to unblock.

Thanks for looking into it

PR #1500, opened 2025-11-10
Clean Up LINQ Tail Recursion Helper

micmarsh:
I found an issue with the usage of `tail` (and the `SelectMany`s that handle `IOTail`, that whole system): if `tail` is the very last call before `Run` everything is fine, but if (as is very likely the case in the field), we call a method that utilizes `tail` internally and proceed with other application logic, the application breaks as the `IOTail` propagates through.

This fixes the issue, as well as consolidates all of the special-case `IOTail` handling logic in one place. 

The first three tests break without the fix (unpacking the `IOTail` once it's handled in `SelectMany`), and the last three are sanity checks to make sure I didn't accidentally make things too permissive. The last three should also break if the internal implementation of LINQ ever changes in such a way that makes the `typeof(B) == typeof(C)` check for unpacking `IOTail` not valid anymore.

### DIFF
--- a/LanguageExt.Core/Effects/IO/DSL/IOTail.cs
+++ b/LanguageExt.Core/Effects/IO/DSL/IOTail.cs
@@ -17,4 +17,12 @@ record IOTail<A>(IO<A> Tail) : IO<A>
 
     public override string ToString() => 
         "IO tail";
+
+    public static IO<C> resolve<A, B, C>(A initialValue, IO<B> bindResult, Func<A, B, C> project)
+        => bindResult switch
+           {
+               IOTail<B> tail when typeof(B) == typeof(C) => (IO<C>)(object)tail.Tail,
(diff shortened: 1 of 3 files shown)


LanguageExt.Core/Traits/Monads/MonadIO/MonadIO.Trait.cs
```diff
@@ -44,17 +44,6 @@ static K<M, A> Maybe.MonadIO<M>.LiftIOMaybe<A>(IO<A> ma) =>
 
     static K<M, A> Maybe.MonadIO<M>.LiftIOMaybe<A>(K<IO, A> ma) => 
         M.LiftIO(ma);
-
-    public static virtual K<M, C> SelectMany<A, B, C>(K<M, A> ma, Func<A, IO<B>> bind, Func<A, B, C> project) =>
```
micmarsh: This (and the below) where not used anywhere, perhaps they were intended to do what `resolve` is now doing?

Issue #1501, opened 2025-11-11
Build errors in C# 14


cthompson92:
Hello,

As it happens, today I attempted to clone the repo and play with it locally to further my own understanding, and when I attempted to build the `language-ext.sln` solution today after installing VS2026, there's a series of build errors:

> 2>------ Build started: Project: LanguageExt.Core, Configuration: Debug Any CPU ------
> 1>  TestBed.Web.Runner -> D:\git\..\Samples\TestBed.Web.Runner\bin\Debug\net8.0\TestBed.Web.Runner.dll
> 2>D:\git\..\LanguageExt.Core\Monads\Alternative Monads\Option\Option.Prelude.cs(90,24,90,29): error CS1503: Argument 1: cannot convert from 'A' to 'System.Action<A>'
> 2>D:\git\..\LanguageExt.Core\Monads\Lifting\Lift.Prelude.cs(92,24,92,28): error CS1061: 'OptionT<IO, A>' does not contain a definition for 'Lift' and no accessible extension method 'Lift' accepting a first argument of type 'OptionT<IO, A>' could be found (are you missing a using directive or an assembly reference?)
> 2>D:\git\..\LanguageExt.Core\Effects\IO\IO.cs(302,23,302,29): error CS0117: 'OptionT<M, A>' does not contain a definition for 'LiftIO'
> 2>D:\git\..\LanguageExt.Core\Effects\IO\IO.cs(306,14,306,26): error CS0314: The type 'M' cannot be used as type parameter 'M' in the generic type or method 'TryT.liftIO<M, A>(IO<A>)'. There is no boxing conversion or type parameter conversion from 'M' to 'LanguageExt.Traits.MonadIO<M>'.
> 2>D:\git\..\LanguageExt.Core\Effects\IO\IO.cs(319,21,319,36): error CS0314: The type 'M' cannot be used as type parameter 'M' in the generic type or method 'ValidationT.liftIO<L, M, A>(IO<A>)'. There is no boxing conversion or type parameter conversion from 'M' to 'LanguageExt.Traits.MonadIO<M>'.
> 2>D:\git\..\LanguageExt.Core\Immutable Collections\Arr\Arr.Module.cs(110,24,110,25): error CS0023: Operator '.' cannot be applied to operand of type 'void'
> 2>D:\git\..\LanguageExt.Core\Immutable Collections\Arr\Arr.cs(605,29,605,30): error CS0023: Operator '.' cannot be applied to operand of type 'void'
[...]

PR #1502, opened 2025-11-12
Fixes for csharp 14 build errors

cthompson92:
Attempt to fix C#14 build errors

As mentioned in #1501, there are two build errors introduced when trying to build with [C#14](https://learn.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-14#implicit-span-conversions) installed.
This PR resolves these issues with an explicit call to `.AsEnumerable()`. Alternative fix could be just to do a cast to `IEnumerable<T>`.

> 2>D:\git..\LanguageExt.Core\Immutable Collections\Arr\Arr.Module.cs(110,24,110,25): error CS0023: Operator '.' cannot be applied to operand of type 'void'
> 2>D:\git..\LanguageExt.Core\Immutable Collections\Arr\Arr.cs(605,29,605,30): error CS0023: Operator '.' cannot be applied to operand of type 'void'

Tested both on C#14 as well as with an explicit `LangVersion 13.0`, this compiles fine on both.

### DIFF
--- a/LanguageExt.Core/Immutable Collections/Arr/Arr.Module.cs
+++ b/LanguageExt.Core/Immutable Collections/Arr/Arr.Module.cs
@@ -107,7 +107,7 @@ public partial class Arr
     /// <returns>Reversed list</returns>
     [Pure]
     public static T[] rev<T>(T[] array) =>
-        array.Reverse().ToArray();
+        array.AsEnumerable().Reverse().ToArray();
 
     /// <summary>
     /// Reverses the array (Reverse in LINQ)
--- a/LanguageExt.Core/Immutable Collections/Arr/Arr.cs
+++ b/LanguageExt.Core/Immutable Collections/Arr/Arr.cs
@@ -602,7 +602,7 @@ public struct Arr<A> :
     [Pure]
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
     public Arr<A> Reverse() =>
-        new (Value.Reverse().ToArray());
+        new (Value.AsEnumerable().Reverse().ToArray());
 
     /// <summary>
     /// Impure iteration of the bound values in the structure


louthy:
There aren't any build errors, C# 14 related or otherwise.  If there are then it's likely to be local issues on your machine related to installed SDKs and the like. 

Feel free to explain why C# 14 causes the errors you see and I'll review, but I'm not going to make random changes without any justification.  Your justifications do not seem to be your own and appear to be from AI tool advice. Please don't make commits based on AI tools.

cthompson92:
Well, I can still reproduce the build errors at the dotnet command line on the latest version of main:
`dotnet build -c Release`

> Restore complete (2.5s)
>   TestBed.Web.Runner net8.0 succeeded (8.4s) → Samples\TestBed.Web.Runner\bin\Debug\net8.0\TestBed.Web.Runner.dll
>   LanguageExt.Core net8.0 failed with 2 error(s) and 1 warning(s) (11.2s)
>     D:\git\hub\louthy\language-ext\LanguageExt.Core\Effects\IO\DSL\IOTail.cs(21,33): warning CS0693: Type parameter 'A' has the same name as the type parameter from outer type 'IOTail<A>'
>     D:\git\hub\louthy\language-ext\LanguageExt.Core\Immutable Collections\Arr\Arr.Module.cs(110,24): error CS0023: Operator '.' cannot be applied to operand of type 'void'
>     D:\git\hub\louthy\language-ext\LanguageExt.Core\Immutable Collections\Arr\Arr.cs(605,29): error CS0023: Operator '.' cannot be applied to operand of type 'void'

For reference, here's my environmental info:

> dotnet --version
> 10.0.100
> dotnet --list-sdks
> 9.0.306 [C:\Program Files\dotnet\sdk]
> 10.0.100 [C:\Program Files\dotnet\sdk]

Though, I am now noticing that that the target framework for all the projects is `net8.0`, and with `LangVersion default`, that _should_ be using C#12, regardless of what is installed. So it is curious that building in VS2026 initially triggered these errors, and that it would happen when using `dotnet build` in this configuration.

This certainly seems more like a tooling bug than an issue with this project - as this _shouldn't_ break anything until the projects actually target net10.

cthompson92:
I did submit this as [a bug in the sdk](https://github.com/dotnet/roslyn/issues/81189#issuecomment-3526997860), though it was closed and the explanation given was that this is how LangVersion default is supposed to work, which is surprising but I guess with a long history of C# not having breaking changes until now, that makes sense. I guess maybe something to be weary of with newer C# versions in the future if they continue with more breaking changes.

PR #1539, opened 2026-02-12
Add Generic Kleisli Composition Operator

micmarsh:
A small thing I've found useful in my local experimentation, seems worthwhile here given how generic it is.

@louthy let me know what you think; I don't have the time or inclination to create versions of this for each concrete data type (wouldn't want to put that on you either!), but I think this should be useful enough as-is


louthy:
Yes, I think it's time, Kleisli works elegantly now. I'm on various skiing trips right now, so I'll follow up when I get back. 

Thanks for the contribution.

PR #1547, opened 2026-04-24
Feature: Domain traits expansion

HernanFAR:
# Domain: Types, Factories & Rules foundation

## Summary

This PR introduces the foundational structure for the Domain layer.

It defines the core building blocks required to model domain logic in a composable, safe, and expressive way:

* **Types**: model domain concepts and meaning
* **Factories**: provide controlled construction of domain values
* **Rules**: enable reusable and composable validation/refinement logic

The goal is to clearly separate *what a value is*, *how it is created*, and *how is validated*, enabling safer and more predictable domain modeling.

This also reinforces key principles:

* no contextless primitives
* explicit error handling (`Fin`, `Either`, etc.)
* composition over inheritance
* no exceptions for flow control
* explicit modeling of effects (`IO`, `Eff`, etc.)

---

## How to use

At this stage, the focus is on defining the building blocks and demonstrating their composition.

* Use **Types** to represent domain concepts (identifiers, magnitudes, refined values, etc.)
* Use **Factories** as the only entry point to construct values
* Use **effectful factories (`DomainFactoryM`)** when creation requires context (time, randomness, IO, etc.)
* Use **Rules** to define reusable validation and refinement logic

A complete domain sample is included to demonstrate:

* refined types built from reusable rules
* effectful modules (Time, Random, Sequences)
* algebraic domain modeling (magnitudes, affine spaces, etc.)
* controlled state evolution via maintainers

---

## Samples
To show the full use of this new traits, we did an update to the DomainTypeSamples sample project. It has an README in english and spanish to allow deep understanding of what we are working with and how we use the traits to go a effective solution

---

## Documentation

Initial README files have been added in both English and Spanish to improve accessibility for a wider audience.

More complete examples and documentation will be added as the framework evolves.


HernanFAR:
I'm developing a big update for the Domain traits, lemme have a time :)

PR #1548, opened 2026-04-25
Feature: Add Maintainer trait, prelude utilities and related free functionality

HernanFAR:
A time ago I create a discution for this very same Maintainer trait. The idea is between now, as I use it more in my proyects, and I think it would bring good utility in the library.


HernanFAR:
Absorved by PR #1547

PR #1549, opened 2026-04-27
Feature: Traits representing domain-invariance: Rule, RuleK, RuleM and RuleT

HernanFAR:
# Rule trait, for Domain invariants!

## Summary
With this traits, the idea is have more composability when representing domain invariants. They are aplicable to any object, agreggate roots value objects, primitive values, etc. We have 4 types currently

- Rule: For invariants of pure values, executed in a direct context. It checks ```A``` values and returns ```Fin<A>``` when validating.
- RuleK: For invariants of higher-kinded values, executed in direct context. It checks ```K<F, A>``` values and return ```Fin<K<F, A>>``` when validating.
- RuleM: For invariants of pure values, executed in monadic ```M``` context. It checks ```A``` values an return ```FinT<M, A>``` when validating.
- RuleT: For invariants of higher-kinded values, executed in monadic transformer ```T``` context. It checks ```K<M, A>``` and ```A``` values and returns ```FinT<T, A>``` when validating.

Each one also has a Module that containts combinators with Rules apply many rules at ones (Through ```All``` and ```Any```), negate a existing rule (Through ```Not```) and Lift a lower-context rule to a monadic context (Through ```Lift```. For ```RuleM``` lifts ```Rule``` and for RuleT lifts ```RuleK```

## How to use

At this moment, we have code-samples in the tests (Go to LanguageExt.Tests/TraitTests/Rule*Tests.cs). But, we are working on a whole Domain-Trait sample using all the upgrades that we are coding!


HernanFAR:
Closed temporally, because of more functionality :)

HernanFAR:
Absorbed by #1547

